PR #43, opened 2021-11-03
Floating node menu

fakob:
* Added a floating node menu
* Moved update type selection from inspector into the floating menu
* Added possibility to update multiple at the same time

Screenshot 2021-11-03 at 21 09 23


src/components/FloatingNodeMenu.tsx
```diff
@@ -0,0 +1,179 @@
...
+    selectedNodes.length === 0 ||
+    selectedNodes[0] === undefined
+  ) {
+    return <div />;
+  }
+
+  // returns null for a specific property,
+  // if its value is not the same throughout the array
+  // else it returns the value
+  const getUpdateBehaviourStateForArray = () => {
```
magnificus, line 34: This is well encapsulated so no big issue but can be made nicer using a reduce() over the nodes and doing the whole struct at once.

src/components/FloatingNodeMenu.tsx
```diff
@@ -0,0 +1,179 @@
...
+import PPNode from '../classes/NodeClass';
+import { ColorWidget } from '../widgets';
+import styles from './../utils/style.module.css';
+
+const FloatingNodeMenu = (props) => {
+  const selectedNodes: PPNode[] = props.selectedNodes;
+  if (
+    selectedNodes === null ||
+    selectedNodes.length === 0 ||
+    selectedNodes[0] === undefined
```
magnificus, line 26: what is the situation where selectedNodes[0] is null? Seems like a dangerous state and might be better to crash if this happens to avoid it propagating.

magnificus (approved):

PR #50, opened 2022-01-06
Switch from blueprintjs to mui

fakob:
I have now moved almost everything from blueprintjs to mui. The one thing missing is the Table node.
I changed the styling a bit, but am still struggling with getting the graph search on top update after editing the name or deleting a graph.
I have also removed dat.gui which was not used anymore and list under the security issues.

Screen Shot 2022-01-06 at 20 59 08
Screen Shot 2022-01-06 at 20 59 35
Screen Shot 2022-01-06 at 20 59 57

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -893,10 +893,13 @@ export default class PPGraph {
 
   deleteSelectedNodes(): void {
     const storedSelection = this.selection.selectedNodes;
+    console.log(storedSelection);
     this.selection.deselectAllNodesAndResetSelection();
-    // loop through storedSelection
-    storedSelection.forEach((node) => {
-      this.removeNode(node);
-    });
+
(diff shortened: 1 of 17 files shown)


src/classes/GraphClass.ts
```diff
@@ -893,10 +893,13 @@ export default class PPGraph {
...
+    console.log(storedSelection);
     this.selection.deselectAllNodesAndResetSelection();
-    // loop through storedSelection
-    storedSelection.forEach((node) => {
-      this.removeNode(node);
-    });
+
+    // loop through storedSelection backwards
+    // as the array I am iterating on gets mutated as well
+    for (let i = storedSelection.length - 1; i >= 0; i--) {
```
magnificus, line 901: could be done nicer with filter + foreach but no biggie

magnificus (approved):

PR #53, opened 2022-01-15
Editor updates

fakob:
I have
* changed the embedded editor to use codemirror (used in codesandbox)
* made an editor node (currently only javascript code highlighting)
* added an accordion for IN, OUT and added a CODE section
  * for custom nodes this CODE view is editable, for normal nodes it is, for now, only viewable
* made the drawer resizable with the mouse or the predefined buttons in the top right

Screen Shot 2022-01-15 at 18 34 12
Screen Shot 2022-01-15 at 18 34 28
Screen Shot 2022-01-15 at 18 29 53


src/widgets.tsx
```diff
@@ -257,6 +257,32 @@ export const TextWidget: React.FunctionComponent<TextWidgetProps> = (props) => {
...
+  const [data, setData] = useState(props.data);
+  console.log(props);
+
+  return (
+    <CodeEditor
+      value={data || ''}
+      randomMainColor={props.randomMainColor}
+      editable={!props.hasLink}
+      onChange={(value) => {
+        console.log('value:', value);
```
magnificus, line 278: unnecessary log

magnificus (approved):

PR #54, opened 2022-01-18
Make JSONKeys and JSONValues node more lenient

fakob:
When reloading the Get example the JSONKeys and JSONValues nodes kept throwing an error and the wiring would not show up. I think it was due to the array lookup getting an undefined or null value in some cases.


fakob:
Just saw that this still ends up firing an error, but at least it was not breaking the building of the graph.
I have now updated the PR and suggest to return an empty array for both `Object.keys` and `Object values` if their input value is undefined or null. What do you think @magnificus ? Or should such for example `Object.keys(undefined)` result in `undefined`?

magnificus (approved):

PR #57, opened 2022-01-30
Json selector node

fakob:
So, this is now my PR where I
* Added a floating socket inspector which can be dragged - to open, just click on any socket name
* Added a json path picker which can be launched from within a json type inspector
* Added option to copy values to the clipboard
* Made the CodeWidget to the standard widget for the any type/AbstractType
* Removed storing the custom property of the socket as it currently is only used when creating nodes.

I know that I would have to do some updates before merging, but I wanted to get it out and have you look at it.

Screen Shot 2022-01-30 at 15 01 28
Screen Shot 2022-01-30 at 09 59 00


src/App.tsx
```diff
@@ -132,6 +132,13 @@ const App = (): JSX.Element => {
   const defaultDrawerWidth = 320;
   const [drawerWidth, setDrawerWidth] = useState(defaultDrawerWidth);
 
+  // socket info
+  const [socketInspectorPosition, setSocketInspectorPosition] =
+    useState<PIXI.Point | null>(null);
```
magnificus, line 137: I would remove possible null state from here, let's say its always a point?

src/App.tsx
```diff
@@ -132,6 +132,13 @@ const App = (): JSX.Element => {
...
+  const [socketToInspect, setSocketToInspect] = useState<PPSocket | undefined>(
```
magnificus, line 138: Also, null state makes sense in this possibility, but it will give us a lot if we don't have this state in App.tsx, that means that whenever this changes the entire App needs a full re-render (this is what was going on with selection), could we maybe get this, selection and floatingmenu as part of their own smaller thing? So they can update without full App.tsx re-render.

src/components/FloatingSocketInspector.tsx
```diff
@@ -0,0 +1,150 @@
...
+            sx={{
+              px: '8px',
+              py: '4px',
+              color: 'text.primary',
+              fontWeight: 'medium',
+              display: 'inline-flex',
+              alignItems: 'center',
+            }}
+          >
+            {props.socketToInspect?.parent?.name}.{props.socketToInspect?.name}
```
magnificus, line 91: Is it possible to launch this menu with no socket in props? it isnt right? menu is not launched from above? If so, remove null checks here.

magnificus: We want to crash if we encounter null here.

src/components/FloatingSocketInspector.tsx
```diff
@@ -0,0 +1,150 @@
...
+          <IconButton size="small" onClick={props.closeSocketInspector}>
+            <CloseIcon sx={{ fontSize: '16px' }} />
+          </IconButton>
+        </Box>
+        <Box id="draggable-content">
+          <PropertyContainer
+            key={0}
+            property={props.socketToInspect}
+            index={0}
+            dataType={props.socketToInspect?.dataType}
```
magnificus, line 136: same with these fellas

src/nodes/data/json.tsx
```diff
@@ -13,7 +15,23 @@ export class JSONGet extends PureNode {
...
+        new StringType(),
+        undefined,
+        undefined,
+        {
+          inspectorInjection: {
+            reactComponent: FloatingJsonPathPicker,
+            props: {
+              jsonSocketName: JSONName,
+              jsonPathSocketName: JSONParamName,
+              forceRefresh: Math.random(),
```
magnificus, line 30: neccessary?

src/widgets.tsx
```diff
@@ -265,15 +267,18 @@ export type CodeWidgetProps = {
 };
 
 export const CodeWidget: React.FunctionComponent<CodeWidgetProps> = (props) => {
-  const [data, setData] = useState(props.data);
+  const [data, setData] = useState(
```
magnificus, line 270: Possible cleaner way to do this is to define the result inside the type classes themselves (so code defines JSON.stringify(props.data, getCircularReplacer(), 2)) and maybe default is just return data

src/widgets.tsx
```diff
@@ -284,20 +289,19 @@ export const CodeWidget: React.FunctionComponent<CodeWidgetProps> = (props) => {
 export const JSONWidget: React.FunctionComponent<TextWidgetProps> = (props) => {
   const [data, setData] = useState(props.data);
   const [displayedString, setDisplayedString] = useState(
-    typeof data === 'object' ? JSON.stringify(data) : data
+    typeof data === 'object'
```
magnificus, line 292: same here, cleaner to use our type system (typeof is usually considered a hack)

magnificus (approved):

src/App.tsx
fakob: Done

src/App.tsx
fakob: Will do in another PR

src/components/FloatingSocketInspector.tsx
fakob: Done

src/components/FloatingSocketInspector.tsx
fakob: Done

src/nodes/data/json.tsx
fakob: Apparently not or not anymore :-)

src/widgets.tsx
fakob: Done

PR #58, opened 2022-02-02
Moved out overlays

fakob:
I have now moved the following overlays out of App.tsx
* Drawer
* Node menu
* Socket inspector

First I wanted to put them all together in one component, but saw that for example the drawer component, which is quite heavy, always rerender when you would move a node around. Now they are separated by changing values.


magnificus (approved):
Very nice!

PR #59, opened 2022-02-03
Log node

fakob:
Added a log viewer node using https://github.com/mozilla-frontend-infra/react-lazylog. It is mainly built for files or url streams. As we are logging mainly incoming text I have created a state with an array. The length of the array can be specified. Not sure if this is the best way to handle that though. Seems to work.

Also made a zindex fix for graph and node search.


src/nodes/logViewer.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    this.addInput('input', new CodeType(), customArgs?.data, true);
+    this.addInput('rowLimit', new NumberType(true, 1, 1000), 100, false);
+
+    this.name = 'LogViewer';
+    this.description = 'View your logs';
+
+    // when the Node is added, add the container and react component
+    this.onNodeAdded = () => {
+      const data = this.getInputData('input');
```
magnificus, line 32: is there reason you need to fetch this data here from the socket directly? very preferable if we can limit that to inside the onExecute function (because we can then make assumptions and optimizations regarding the execution flow. Execute should be called _onAdded so this might just be running twice

magnificus (approved):
Approved but added one comment

src/nodes/logViewer.tsx
fakob: You are right. Fixed.

PR #60, opened 2022-02-05
purenode and label change

magnificus:


src/nodes/text.tsx
```diff
@@ -950,7 +459,8 @@ export class Note extends PPNode {
         }
         this.fontSize = this._bitmapTextRef.fontSize;
         this._bitmapTextRef.text = data;
-        this.setOutputData('data', data);
+        //this.setOutputData('data', data);
+        this.executeOptimizedChain();
       }
```
fakob, line 464: If setting the output data is removed, I assume that the output data will not be updated. Or am I missing something?

fakob (approved):

src/nodes/text.tsx
magnificus: Ah hmm I see there is no real execute function in this class, my misunderstanding, will change.

PR #61, opened 2022-02-05
Fix slice issue

fakob:


src/nodes/datatypes/abstractType.tsx
```diff
@@ -19,8 +19,11 @@ export class AbstractType {
     return 'null';
   }
 
-  getInputWidget = (data: any): any => {
-    return <CodeWidget {...data} />;
+  getInputWidget = (props: any): any => {
```
magnificus, line 22: Ah hmm this is quite problematic as we want to be able to support data types that are not strings, this is in the top class, this is good for the conversion logic, but it should be in the codetype/stringtype/etc respectively. I can (or you can if you want) change this in the sub classes respectively instead?

magnificus (changes requested):

src/nodes/datatypes/abstractType.tsx
magnificus: ah ok youre using codewidget as the default widget, ok we can do that

magnificus (approved):

PR #62, opened 2022-02-09
Deferred drawing workflow

magnificus:
New pixi drawing, we can walk through it and discuss before any merging, this definitely risks stepping on toes (and will break old examples)


src/nodes/draw2.tsx
```diff
@@ -0,0 +1,433 @@
...
+const scaleYName = 'Scale Y';
+const inputRotationName = 'Rotation';
+
+const inputShapeName = 'Shape';
+const inputColorName = 'Color';
+const inputSizeName = 'Size';
+const inputBorderName = 'Border';
+const outputPixiName = 'Graphics';
+
+const inputCombine1Name = 'Primary';
```
fakob, line 51: Maybe call it foreground and background?

src/nodes/draw.tsx
```diff
@@ -10,827 +10,10 @@ import * as csvParser from 'papaparse';
...
-    this.addInput('angle', new NumberType(true, -360, 360), 0);
-    this.addInput(
-      'align',
-      new EnumType(PIXI_TEXT_ALIGN_OPTIONS),
-      'left',
-      false
-    );
-    this.addInput('pivot', new EnumType(PIXI_PIVOT_OPTIONS), 'top left', false);
-    this.addInput('size', new NumberType(true, 1), 24, undefined);
-    this.addInput('color', new ColorType(), hexToTRgba(fillColor));
```
fakob, line 61: For the text node could you add text color and align as inputs?

fakob: I believe it might also need a width parameter for text wrap to work properly?

src/nodes/draw.tsx
```diff
@@ -10,827 +10,10 @@ import * as csvParser from 'papaparse';
...
-
-    super(name, graph, {
-      ...customArgs,
-      color: nodeColor,
-    });
-
-    this.addOutput('graphics', new AnyType());
-    this.addInput('x', new NumberType(), 0);
-    this.addInput('y', new NumberType(), 0);
-    this.addInput('pivot', new EnumType(PIXI_PIVOT_OPTIONS), 'top left', false);
```
fakob, line 315: Being able to set the pivot point can be very helpful. Could we add that as well? I think it would be valuable for most/all of the nodes.

fakob:
Let me know when you merge it I can then update the examples.

fakob (approved):

fakob:
Forgot one thing. Could you not delete the old PIXI nodes? This would help for recreating the examples, I can delete them afterwards. I believe you have new node names anyway.

src/nodes/draw.tsx
```diff
@@ -10,827 +10,10 @@ import * as csvParser from 'papaparse';
...
-      const x = [].concat(input['x']);
-      const y = [].concat(input['y']);
-      const width = [].concat(input['width']);
-      const angle = [].concat(input['angle']);
-      const text = [].concat(input['text']);
-      const size = [].concat(input['size']);
-      const color = [].concat(input['color']);
-      const align = input['align'];
-      const pivot = input['pivot'];
-      const lengthOfLargestArray = Math.max(
```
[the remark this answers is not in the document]
magnificus: No that's not possible ATM, IMO should cut as it is a pretty clean separation of concerns

PR #63, opened 2022-02-13
Improve image handling and drag and drop

fakob:
I updated the image node with the following features
* imageType now shows a preview of the image
  * the browsers right click menu is now only disabled over the canvas, so right clicking on html images to save the image is possible
* resizing the image now works
  * shift resizing nodes, now constrains their aspect ratio
* added css like image fit input option
* added a reset size button
* added image details output
* fixed issue that we did not save the defaultData (backup of data if a link is plugged into an input socket)

Regarding drag and drop I have made the following changes
* nodes are now created at the drop position
* newly created nodes are selected and zoomed into
* multiple images can be dragged in at the same time and would arrange in a grid
* if one or more image nodes are previously selected, the dragged in images would overwrite these with the new images

Also
* disabled the history back gesture which allowed multi finger gestures on trackpads to go backwards/forwards in history
* added local and global node bounds to the debug output text


src/classes/NodeClass.ts
```diff
@@ -150,11 +155,16 @@ export default class PPNode extends PIXI.Container {
...
       }),
     ];
+    const debugText = new PIXI.Text('', COMMENT_TEXTSTYLE);
+    debugText.resolution = 1;
     const nodeComment = new PIXI.Text('', COMMENT_TEXTSTYLE);
     nodeComment.resolution = 1;
 
     this._BackgroundRef = this.addChild(background);
     this._NodeNameRef = this.addChild(inputNameText);
+    this._NodeDebugRef = (
```
magnificus, line 165: do commentcontainer = this.graph.viewport.getChildByName('commentContainer'); and then use it twice to avoid copy pasting

src/classes/NodeClass.ts
```diff
@@ -548,10 +568,28 @@ export default class PPNode extends PIXI.Container {
...
 
-  resizeNode(width: number, height: number): void {
+  resizeNode(width: number, height: number, maintainAspectRatio = false): void {
     // set new size
-    this.nodeWidth = Math.max(width, this.minNodeWidth);
-    this.nodeHeight = Math.max(height, this.calculatedMinNodeHeight);
+    let newNodeWidth = Math.max(width, this.minNodeWidth);
+    let newNodeHeight = Math.max(height, this.calculatedMinNodeHeight);
+
+    if (maintainAspectRatio) {
```
magnificus, line 576: easier to follow if this is just an if() else()

magnificus (approved):
Looks good added some comments

src/App.tsx
```diff
@@ -127,20 +131,29 @@ const App = (): JSX.Element => {
...
+      new PIXI.Point(event.clientX, event.clientY)
+    );
+    let nodePosX = dropPoint.x;
+    const nodePosY = dropPoint.y;
+    const newNodeSelection: PPNode[] = [];
+
+    (async function () {
+      for (let index = 0; index < acceptedFiles.length; index++) {
+        const file = acceptedFiles[index];
+        console.log(file);
```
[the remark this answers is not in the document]
fakob: removed them

src/classes/NodeClass.ts
fakob: done

src/classes/NodeClass.ts
fakob: done

src/App.tsx
```diff
@@ -508,19 +561,35 @@ const App = (): JSX.Element => {
...
     );
-    viewport.current.moveCenter(
-      nodeContainerBounds.x + nodeContainerBounds.width / 2,
-      nodeContainerBounds.y + nodeContainerBounds.height / 2
-    );
-    viewport.current.zoomPercent(-0.1, true); // zoom out a bit
+    viewport.current.zoomPercent(-0.2, true); // zoom out a bit
+  };
+
+  const zoomToFitSelection = () => {
```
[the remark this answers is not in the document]
fakob: done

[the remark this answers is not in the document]
fakob: and added a menu entry in the node selection

PR #65, opened 2022-02-19
Color refactor

magnificus:
made us use TRgba as a base everywhere, made it a class and gave it functionality


src/classes/SocketClass.ts
```diff
@@ -211,7 +211,9 @@ export default class Socket extends PIXI.Container {
     // visibility change can result in position change
     // therefore redraw Node and connected Links
     this.getNode().drawNodeShape();
-    this.getNode().inputSocketArray.map((input) => {
+
+    // is below neccessary? if so should be inside node and not here
+    /*this.getNode().inputSocketArray.map((input) => {
       input.links.map((link) => {
```
fakob, line 217: I believe this redraws the connecting lines. If you hide a socket further up, all other sockets move up and therefore the connections have to be redrawn. I agree, this should be a separate function in the NodeClass, which is called from here.

src/classes/SocketClass.ts
```diff
@@ -51,12 +49,10 @@ export default class Socket extends PIXI.Container {
 
     let defaultData;
     if (socketType === SOCKET_TYPE.IN) {
-      // define defaultValue for different types
-      if (data === null && dataType && dataType instanceof AbstractType) {
+      // define defaultData for different types
+      if (data === null && dataType) {
```
fakob, line 53: Haha, I had renamed this to defaultValue as we are using the term defaultData for the "data backup" if a socket has a link. But it probably does not matter.

src/nodes/datatypes/pixiType.tsx
```diff
@@ -1,6 +1,7 @@
 import { inspect } from 'util';
 import { AbstractType } from './abstractType';
 
+// Can we deprecate this type once draw.tsx is gone?
 export class PixiType extends AbstractType {
```
fakob, line 5: Yes, I believe we can.

fakob: Which reminds me of, that I have not recreated all of them yet.

fakob (approved):

fakob (approved):

PR #67, opened 2022-02-20
Store viewport position

fakob:
Added saving and loading of viewport position and scale. So after loading you see the exact same section of the graph you had when saving.
For this I have also added a bit more debugging info. Mouse and viewport coordinates on the top left and a grid shader texture spanning from 0,0 - 1000,1000.

Screenshot 2022-02-20 at 17 05 54

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -684,10 +684,14 @@ export default class PPGraph {
     );
 
     const data = {
-      customNodeTypes: this.customNodeTypes,
+      version: PP_VERSION,
+      graphSettings: {
+        viewportCenterPosition: this.viewport.center,
+        viewportScale: this.viewport.scale.x,
+      },
       nodes: nodesSerialized,
       links: linksSerialized,
-      version: PP_VERSION,
+      customNodeTypes: this.customNodeTypes,
     };
 
     return data;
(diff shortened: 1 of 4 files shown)


src/classes/GraphClass.ts
```diff
@@ -715,6 +719,16 @@ export default class PPGraph {
     // store customNodeTypes
     this.customNodeTypes = data.customNodeTypes;
 
+    // position and scale viewport
+    const newX = data.graphSettings?.viewportCenterPosition.x ?? 0;
```
magnificus, line 723: I would remove all "?"'s

src/classes/GraphClass.ts
```diff
@@ -715,6 +719,16 @@ export default class PPGraph {
...
+    const newY = data.graphSettings?.viewportCenterPosition.y ?? 0;
+    this.viewport.animate({
+      position: new PIXI.Point(newX, newY),
+      scale: data.graphSettings?.viewportScale ?? 1,
```
magnificus, line 727: same thing here, avoid null check

magnificus (approved):

src/classes/GraphClass.ts
fakob: I had added the null checks so previously saved graphs would not crash and could be opened. Well, I guess we would have to update all of them anyway :-)

PR #68, opened 2022-02-21
Fix text width

fakob:
Small fix for the text. The width parameter would scale the text, the one we had before is wordWrapWidth in style.

### DIFF
--- a/src/nodes/draw2.tsx
+++ b/src/nodes/draw2.tsx
@@ -308,8 +308,7 @@ export class DRAW_Text extends DRAW_Base {
       lineJoin: 'round',
     });
     const basicText = new PIXI.Text(inputObject[inputTextName], textStyle);
-    if (inputObject[inputWidthName] !== 0)
-      basicText.style.fill = inputObject[inputColorName].hex();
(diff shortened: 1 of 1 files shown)


src/nodes/draw2.tsx
```diff
@@ -304,12 +304,12 @@ export class DRAW_Text extends DRAW_Base {
...
       whiteSpace: 'pre-line',
       wordWrap: true,
+      wordWrapWidth: inputObject[inputWidthName],
       lineJoin: 'round',
     });
     const basicText = new PIXI.Text(inputObject[inputTextName], textStyle);
     if (inputObject[inputWidthName] !== 0)
-      basicText.width = inputObject[inputWidthName];
-    basicText.style.fill = inputObject[inputColorName].hex();
+      basicText.style.fill = inputObject[inputColorName].hex();
```
magnificus, line 312: now it will only take effect if (inputObject[inputWidthName] !== 0), i would remove line 311, otherwise approved :)

fakob: I was not sure why `if (inputObject[inputWidthName] !== 0)` was in there. That's why I had not touched it. Ok, then I will remove it.

PR #69, opened 2022-02-21
Math node

fakob:
Added some math related nodes:
* A node which includes all operators and functions of Math. The display name changes according to the selected option
* Separate nodes for add, subtract, multiply and divide

I also added
* a number output widget and a conversion to number
* and allowed changing the type on the output
This way you can potentially do conversion, for example get a number out of the Label

Screen Shot 2022-02-21 at 22 43 36

### DIFF
--- a/src/nodes/datatypes/numberType.tsx
+++ b/src/nodes/datatypes/numberType.tsx
@@ -1,6 +1,10 @@
 import React from 'react';
 import { TRgba } from '../../utils/interfaces';
-import { SliderWidget, SliderWidgetProps } from '../../widgets';
+import {
+  NumberOutputWidget,
+  SliderWidget,
+  SliderWidgetProps,
+} from '../../widgets';
 import { AbstractType } from './abstractType';
 
 export class NumberType extends AbstractType {
@@ -33,6 +37,13 @@ export class NumberType extends AbstractType {
     return <SliderWidget {...sliderProps} />;
   };
 
+  getOutputWidget = (props: any): any => {
+    if (typeof props.data !== 'number') {
+      props.data = Number(props.data);
+    }
+    return <NumberOutputWidget {...props} />;
+  };
+
   getName(): string {
     return 'Number';
   }
@@ -43,5 +54,9 @@ export class NumberType extends AbstractType {
 
   getColor(): TRgba {
     return new TRgba(128, 229, 229);
+  }
+
+  parse(data: any): any {
+    return Number(data);
   }
 }
--- a/src/nodes/math.ts
+++ b/src/nodes/math.ts
@@ -88,94 +88,125 @@ export class MathFunction extends PureNode {
   }
 }
 
-export class Add extends PureNode {
+class SimpleMathOperation extends PureNode {
   constructor(name: string, graph: PPGraph, customArgs: CustomArgs) {
     super(name, graph, {
       ...customArgs,
       color: TRgba.fromString(NODE_TYPE_COLOR.TRANSFORM),
     });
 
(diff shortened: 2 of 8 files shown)


fakob:
Made an additional update: Moved renaming of node out of the execute function into an onOptionChange callback which I added to the enum type. This way we can listen to updates independently.

src/nodes/datatypes/numberType.tsx
```diff
@@ -33,6 +37,13 @@ export class NumberType extends AbstractType {
     return <SliderWidget {...sliderProps} />;
   };
 
+  getOutputWidget = (props: any): any => {
+    if (typeof props.data !== 'number') {
```
magnificus, line 41: I would use the sanitization step to fix this rather than checking data manually, but i can do that.

src/nodes/math.ts
```diff
@@ -3,108 +3,163 @@ import PPNode from '../classes/NodeClass';
...
+      'SQRT2',
+    ];
+    const staticMethodsWith0Parameters = ['random'];
+    const staticMethodsWith2Parameters = ['atan2', 'imul', 'pow'];
+
+    this.onOptionChange = (value) => {
+      this.nodeName = 'Math.' + value;
+    };
+
+    this.addInput('Input', new NumberType(false, -10, 10), 0, true);
```
magnificus, line 41: could we add these via getDefaultIO function? it is cleaner than adding them in constructor

src/nodes/math.ts
```diff
@@ -3,108 +3,163 @@ import PPNode from '../classes/NodeClass';
 import { NODE_TYPE_COLOR } from '../utils/constants';
 import { CustomArgs, TRgba } from '../utils/interfaces';
 import { NumberType } from './datatypes/numberType';
+import { EnumType } from './datatypes/enumType';
 
-export class MathAdd extends PPNode {
+export class MathFunction extends PPNode {
```
magnificus, line 8: should definitely be pure node

src/nodes/math.ts
```diff
@@ -3,108 +3,163 @@ import PPNode from '../classes/NodeClass';
...
+    this.onExecute = async function (input, output) {
+      const a = input['Input'];
+      const b = input['Input2'];
+      const result = a - b;
+      output['Output'] = result;
     };
+  }
+}
+
+export class Multiply extends PPNode {
```
magnificus, line 121: these nodes have way to much copy pasted code, they should be oneliners, make a node that is called "MathTwoElementOperation" or something with this logic that you are overriding, you should be able to make every node something like 1 single line.

magnificus (changes requested):
I like the new "math function" node a lot, but there are some code issues here we need to fix (copy pasting)

src/nodes/math.ts
magnificus: Tell me if you want me to checkout the branch and change this if you dont know what i am talking about

src/nodes/math.ts
fakob: Done, still not sure what a PureNode is though :-) One without side effects?

src/nodes/math.ts
fakob: Please have a jab at it.

src/nodes/math.ts
magnificus: Yes, one where you say "if there is no difference in the result i generated in the function, then i dont need to tell any other nodes that i have run at all, im guaranteeing i dont have side effects"

src/nodes/math.ts
magnificus: OK

PR #70, opened 2022-02-23
Right clicking a socket will remove the link

fakob:


src/classes/SocketClass.ts
```diff
@@ -296,7 +296,14 @@ export default class Socket extends PIXI.Container {
...
     console.log('_onPointerDown');
     if (event.target.name === 'SocketRef') {
-      this.getGraph().clickedSocketRef = event.target.parent as Socket;
+      if (event.data.button === 2) {
+        this.getGraph().checkIfSocketHasConnectionAndDeleteIt(
+          this,
+          this.isInput()
+        );
+      } else {
+        this.getGraph().clickedSocketRef = event.target.parent as Socket;
```
magnificus, line 305: this is different from line 308, intentional?

magnificus (approved):

src/classes/SocketClass.ts
fakob: Yes, the first is checking a left or right click on the circle of the socket. The second one is checking a left click on the socket name. This one brings up the socket inspector.

PR #75, opened 2022-03-03
More lenient json handling

fakob:
* added the JSON5 package as it is more lenient when it comes to JSON parsing
* unified the JSON parsing in a function catering for both strings and objects as inputs

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -429,26 +429,19 @@ export const replacePartOfObject = (
   return obj;
 };
 
-export const parseJSON = (
-  jsonToParse: any
-): { [key: string]: any } | undefined => {
+export const parseJSON = (jsonToParse: any): { [key: string]: any } => {
   let jsonObj: any;
-  try {
-    switch (typeof jsonToParse) {
-      case 'string':
-        jsonObj = JSON5.parse(jsonToParse);
-        break;
-      case 'object':
-        jsonObj = jsonToParse;
-        break;
+  switch (typeof jsonToParse) {
+    case 'string':
+      jsonObj = JSON5.parse(jsonToParse);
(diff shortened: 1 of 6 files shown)


magnificus (commented):
I would probably prefer letting the parsing fail, as it will generate the red outline on the node, whereas now we won't, can we make it throw the exception instead?

src/utils/utils.ts
```diff
@@ -426,3 +428,27 @@ export const replacePartOfObject = (
   movingPointer[last] = objValue;
   return obj;
 };
+
+export const parseJSON = (
+  jsonToParse: any
+): { [key: string]: any } | undefined => {
+  let jsonObj: any;
+  try {
+    switch (typeof jsonToParse) {
```
fakob, line 437: @magnificus do you mean to remove try/catch here? As it will be caught in the node execute? 
Currently I am using this function also in a none node context I guess. For example in the JSONPathPicker. Should I then make 2 different functions? One which uses a try catch and one without?

magnificus: I see, would it be possible to catch it up in JSONPathPicker? then we can let it slide through here, and JSONPathPicker will intercept, and otherwise we will let it through, and we won't need two functions.

magnificus: also nice to get rid of undefined

magnificus: If you want to contain the error here then would recommend at least not returning undefined (so for example returning {} if something goes wrong).

fakob: I have now moved try catch out of parseJSON function and into the 2 places where it was missing them. Now the node red border gets nicely triggered.

magnificus (approved):

PR #76, opened 2022-03-05
Fixing floating sockets

fakob:
Reverting this small change here:
https://github.com/fakob/plug-and-play/commit/f4dda942e1152f4eb8aa0522c5b5181671965ba0#r68035206


fakob:
Accidentally merged this, but pushed now a fix directly to main using filter

PR #79, opened 2022-03-18
Save presets

fakob:
For getting save presets to work I have added a couple of nodes and made some other changes:
* introduced a TriggerNode class - interestingly enough does the socket sometimes be added on the top, sometimes after all native sockets. not sure why
* changed some of the nodes to inherit from the TriggerNode class
* added Keyboard node - gets keyboard events
* added TriggerOnUpFlank node - create trigger on value change from 0 to 1, includes a delay option
* added PassThrough node - pass input to output on trigger
* added If_Else node
* added Compare node - comparing 2 values
* added Check node - checking if an input is undefined or null
* updated Console node to also print the node id
* updated the Multiplier node to expose onPointerDown

Here the current save presets graph to test with:
[preset test 05 - 2022-03-18 at 20.32.23.ppgraph.zip](https://github.com/fakob/plug-and-play/files/8307231/preset.test.05.-.2022-03-18.at.20.32.23.ppgraph.zip)

Screen_Shot_2022-03-16_at_19 55 31

### DIFF
--- a/src/nodes/base.ts
+++ b/src/nodes/base.ts
@@ -15,7 +15,12 @@ import {
   SOCKET_TYPE,
 } from '../utils/constants';
 import { CustomArgs, TRgba } from '../utils/interfaces';
-import { compare, getMethods, isInputTrue } from '../utils/utils';
+import {
+  compare,
+  convertToString,
+  getMethods,
+  isInputTrue,
+} from '../utils/utils';
 import { NumberType } from './datatypes/numberType';
 import { AnyType } from './datatypes/anyType';
 import { TriggerType } from './datatypes/triggerType';
@@ -23,6 +28,7 @@ import { ColorType } from './datatypes/colorType';
 import { StringType } from './datatypes/stringType';
 import { EnumType } from './datatypes/enumType';
 import { BooleanType } from './datatypes/booleanType';
+import { JSONType } from './datatypes/jsonType';
 
 export class Mouse extends PPNode {
   onViewportMove: (event: PIXI.InteractionEvent) => void;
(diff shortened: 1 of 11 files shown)


src/nodes/base.ts
```diff
@@ -268,9 +339,186 @@ export class DateAndTime extends PPNode {
...
+    outputObject: Record<string, unknown>
+  ): Promise<void> {
+    const inputA = inputObject['A'];
+    const inputB = inputObject['B'];
+    const operator = inputObject['Operator'];
+    outputObject['Output'] = compare(inputA, operator, inputB);
+  }
+}
+
+export class Check extends PPNode {
```
magnificus, line 487: Maybe name it something like IsValid?

src/utils/utils.ts
```diff
@@ -445,3 +447,78 @@ export const parseJSON = (jsonToParse: any): { [key: string]: any } => {
   }
   return jsonObj;
 };
+
+export const isInputTrue = (input: unknown): boolean => {
```
magnificus, line 451: use this instead:

https://www.geeksforgeeks.org/lodash-_-truthy-method/

src/nodes/base.ts
```diff
@@ -268,9 +339,186 @@ export class DateAndTime extends PPNode {
...
+        setTimeout(() => {
+          this.trigger();
+        }, delay);
+      }
+      this.previousValue = newValue;
+    };
+  }
+
+  trigger(): void {
+    this.outputSocketArray[0].links.forEach((link) => {
```
magnificus, line 379: same as above, this looks very sus, I think there are better ways to do the cascading here, lets talk about it?

magnificus (changes requested):
Looks very cool overall, just some minor things that we probably want to address

src/nodes/base.ts
```diff
@@ -268,9 +339,186 @@ export class DateAndTime extends PPNode {
...
+      }
+      if ((link.target.parent as any).trigger) {
+        (link.target.parent as any).trigger();
+      }
+    });
+    this.executeChildren();
+  }
+}
+
+export class PassThrough extends TriggerNode {
```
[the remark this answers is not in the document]
fakob: Not sure I understand what you mean with reroute-node, this one is as generic as it can get :-)
This node should pass data only when triggered instead of normal execution.

src/nodes/base.ts
fakob: You are right. I think I want it to be more open to maybe also check for stuff like typeof, but maybe that should be a separate node then.

src/utils/utils.ts
fakob: I can do that, but I would have to install lodash-contrib extra for this.

src/nodes/base.ts
```diff
@@ -222,18 +285,26 @@ export class Trigger extends PPNode {
       this._rectRef.on('pointerdown', this.trigger.bind(this));
     };
   }
+
   trigger(): void {
-    console.log('Triggered node: ', this.name);
     this.outputSocketArray[0].links.forEach((link) => {
-      (link.target.parent as any).trigger();
+      // trigger both a normal trigger as well as a node update
+      if ((link.target.parent as any).triggerUpdate) {
```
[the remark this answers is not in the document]
fakob: We had a `trigger` function from before which get's called when using a triggerType button. This executes a local trigger function which could be used to execute something locally.
In addition I have introduced the `triggerUpdate` function which is the one present on all TriggerNodes. 
I guess they have different responsibilities now. So this does not make sense anymore. 
I probably need to make the triggerType to only trigger local functions unrelated to execution. Let's discuss this.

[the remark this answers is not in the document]
magnificus: (y)

src/nodes/base.ts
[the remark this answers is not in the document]
magnificus: aha ok didnt understand it properly

src/utils/utils.ts
magnificus: Yes, I think its worthwhile, as this will be used to parse user input, and they could do a lot of things that are hard to imagine (so library hopefully will have thought more about possibilities)

magnificus (approved):

PR #81, opened 2022-03-19
Fix duplicateSelection losing links

fakob:
I saw that the duplicateSelection did not work anymore as it was not duplicating the links. I have now removed the async await from the loop which creates the nodes as the nodes are needed before we can start creating the new links.


magnificus (approved):
looks a bit dangerous but assuming it has been tested

PR #82, opened 2022-03-19
Copy paste nodes

fakob:
Added copy paste functionality


magnificus (approved):
This is very useful functionality, can we make the duplicate function now use this under the hood? So we don't define the functionality twice.

fakob:
Good point. Will do.

PR #85, opened 2022-03-21
Improve pan zoom handling

fakob:
Something different I worked on the weekend. I tried to make the pan and zoom handling more similar to Miro/Blueprints and it seems to work well for me. I tested it on a Mac and a Pc with trackpad and mouse. It might need some more testing on different devices. This is the new behaviour:
* Drag with trackpad or middle or right mouse click
* Zoom with trackpad using pinch or ctrl scroll
* Zoom with mouse using scrollwheel
* Select with left click (shift key is only need for adding or subtracting the selection)

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -36,7 +36,7 @@ export default class PPGraph {
   selectedSourceSocket: null | Socket;
   lastSelectedSocketWasInput = false;
   overInputRef: null | Socket;
-  dragSourcePoint: null | PIXI.Point;
+  dragSourcePoint: PIXI.Point;
 
   backgroundTempContainer: PIXI.Container;
   backgroundCanvas: PIXI.Container;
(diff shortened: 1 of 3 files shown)


src/classes/SelectionClass.ts
```diff
@@ -303,6 +303,21 @@ export default class PPSelection extends PIXI.Container {
       this.resetAllGraphics();
     }
     this.onSelectionChange(this.selectedNodes);
+
+    // only trigger deselect if the mouse was not moved and onMove was not called
+    const targetPoint = new PIXI.Point(
+      (event.data.originalEvent as MouseEvent).clientX,
+      (event.data.originalEvent as MouseEvent).clientY
+    );
+    if (this.sourcePoint !== null) {
```
magnificus, line 312: Avoid null check (lets instead never let it be null)

src/classes/GraphClass.ts
```diff
@@ -141,8 +141,14 @@ export default class PPGraph {
     event.stopPropagation();
     const target = event.target;
     console.log(target, event.data.originalEvent);
-
-    this.onRightClick(event, target);
+    if (
+      // only trigger right click if viewport was not dragged
+      this.dragSourcePoint === null ||
```
magnificus, line 146: Remove null check here as well

magnificus (approved):
Looks good but let's remove null checks

PR #87, opened 2022-03-23
Fixed issue when deleting the current graph

fakob:
This fixes an issue where deleting the current graph would lead to an undefined state
And the fallback graph is now the last saved graph

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -726,17 +726,16 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
     console.log(graphId);
     db.transaction('rw', db.graphs, db.settings, async () => {
       const loadedGraphId = await getLoadedGraphId(db);
-      if (loadedGraphId === graphId) {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -800,14 +799,14 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
+      if (graphs.length > 0) {
         let loadedGraph = graphs.find(
           (graph) => graph.id === (id || loadedGraphId)
         );
 
-        // check if graph exists and load last graph if it does not
+        // check if graph exists and load last saved graph if it does not
         if (loadedGraph === undefined) {
-          loadedGraph = graphs[graphs.length - 1];
+          loadedGraph = graphs.reduce((a, b) => (a.date > b.date ? a : b));
```
magnificus, line 809: lol interesting solution

magnificus (approved):

PR #90, opened 2022-04-02
Redone table node

fakob:
As the old table node was never working properly I have now redone it completely. 
The format handling is done using: https://github.com/SheetJS/sheetjs
The spreadsheet editor is done using: https://github.com/myliang/x-spreadsheet. This package was the best one I found, but it has some flaws. Let's see how far we get.

The table node has the following features: 
* Import (drag and drop) of a lot of spreadsheet formats
* Create empty spreadsheet and copy paste data in
* Support for multiple sheets including create/rename/delete
* Support for copy/paste
* Support for adding/removing lines
* Support for editing values
* Outputs
  * the whole workbook - native format
  * the selected worksheet
    * native format
    * json (header is used as keys)
    * array of arrays
    * csv

Known limitations:
* If you want to use another than the first sheet, you can just click on it to select it or use the sheetIndex. However when exiting the table it will always only show the first sheet, even though the data comes from another sheet
* When resizing the width of the table node, the columns are cut off. One needs to, save it and reload the graph to not cut off the columns
* When using ctrl+a to select all cells, also all nodes are selected
* not yet supported
  * Output only selected cells
  * URLs like google spreadsheet
  * Exporting of data

I also changed
* the behaviour of hybrid nodes. they need to be double clicked, as before, but now they active one gets a drop shadow
* the behaviour when dragging in files. We now only ensure that the dragged in nodes are in view and only zoom out if they would cover the hole screen. This is now also applied to duplicate and copy paste. 

Screenshot 2022-04-02 at 16 34 46
Screenshot 2022-04-02 at 16 32 42
Screenshot 2022-04-02 at 16 32 55
Screenshot 2022-04-02 at 16 32 59
Screenshot 2022-04-02 at 16 33 09
Screenshot 2022-04-02 at 16 33 33
Screenshot 2022-04-02 at 16 35 00

### DIFF
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -716,7 +716,7 @@ export default class PPNode extends PIXI.Container {
       if (commentData && commentData.length > 10000) {
         commentData = 'Too long to display';
       }
-      console.log('drawing comments');
+      // console.log('drawing comments');
       const debugText = new PIXI.Text(
         `${Math.round(this.transform.position.x)}, ${Math.round(
           this.transform.position.y
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -106,6 +106,7 @@ export default class PPNode extends PIXI.Container {
   onViewportMoveHandler: (event?: PIXI.InteractionEvent) => void = () => {};
   onViewportPointerUpHandler: (event?: PIXI.InteractionEvent) => void =
     () => {};
+  onHybridNodeExit: () => void = () => {}; // called when a hybrid node is exited after double click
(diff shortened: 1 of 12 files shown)


src/classes/NodeClass.ts
```diff
@@ -1189,14 +1187,33 @@ export default class PPNode extends PIXI.Container {
 
     // turn on pointer events for hybrid nodes so the react components become reactive
     if (this.isHybrid) {
```
magnificus, line 1189: we should make hybridnode a child to node

magnificus (approved):
very cool looking forward to trying it otu

PR #91, opened 2022-04-03
Fix scroll wheel behaviour

fakob:
I have now added a gesture mode. By default it is set to auto-detect the pointer device. In case it detects a mouse/scroll-wheel it sets the scroll wheel to be zoom. This detection happens if you have not manually set to mouse or trackpad mode and happens only on loading.
The mode can be set in the context menu.

Screenshot 2022-04-03 at 12 26 40

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -119,7 +119,6 @@ const App = (): JSX.Element => {
   const overlayCommentContainer = useRef<PIXI.Container | null>(null);
   const graphSearchInput = useRef<HTMLInputElement | null>(null);
   const nodeSearchInput = useRef<HTMLInputElement | null>(null);
-  const [isTrackpad, setIsTrackpad] = useState(isMac);
   const [isGraphSearchOpen, setIsGraphSearchOpen] = useState(false);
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -662,9 +666,70 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
+    window.removeEventListener('DOMMouseScroll', detectTrackPad);
+  }
+
+  function applyGestureMode(viewport: Viewport, newGestureMode = undefined) {
+    db.transaction('rw', db.settings, async () => {
+      let gestureMode = newGestureMode;
+      if (gestureMode) {
+        // save newGestureMode
+        await db.settings.put({
+          name: 'gestureMode',
```
magnificus, line 705: Are we sure we want gesture mode saved with a specific graph?

magnificus (approved):

src/App.tsx
fakob: Don't worry, this is saved next to graphs as "machine settings".
Screen Shot 2022-04-09 at 11 56 21

PR #98, opened 2022-04-21
Macro UNFINISHED but ready to merge

magnificus:
want to merge in already because there are plenty of bugfixes in here and i dont want to diverge too far from main, more work remains on macro


src/nodes/macro/macro.ts
```diff
@@ -0,0 +1,145 @@
+import PPGraph from '../../classes/GraphClass';
+import PPNode, { PureNode } from '../../classes/NodeClass';
+import Socket from '../../classes/SocketClass';
+import { NODE_TYPE_COLOR, SOCKET_TYPE } from '../../utils/constants';
+import { CustomArgs, TRgba } from '../../utils/interfaces';
+import { AnyType } from '../datatypes/anyType';
+import { MacroType } from '../datatypes/macroType';
+
+export class MacroNode extends PPNode {
+  public addDefaultInput(): void {
```
fakob, line 10: Is this just a base class or does it need to be exported which means it shows up in the nodes list?

fakob (approved):

src/nodes/macro/macro.ts
magnificus: ah right does not need to be exported

PR #99, opened 2022-04-21
New trigger input

fakob:
This PR is not done yet, just wanted to show where I am at.
So I think I did what you suggested, but the Socket method `set data(newData: any)` is not called on the Trigger 1 socket when I change the Constant In value. Will have a look on the weekend, where my onSetData method should be hooked into so it gets triggered.

Screen Shot 2022-04-21 at 22 15 20


src/classes/SocketClass.ts
```diff
@@ -143,6 +146,18 @@ export default class Socket extends PIXI.Container {
     this._links = newLink;
   }
 
+  get previousData(): any {
```
magnificus, line 149: avoid copy pasting by merging this and getData

magnificus (approved):
looks ait added one comment

fakob:
Please have another look. 
I had to restrict to one trigger per node, as I did not get multiple triggers on one node to work properly.
With this task I saw my limitations in this area :-(  I think it works okayish now, but it is far from perfect.

src/classes/NodeClass.ts
```diff
@@ -1042,16 +1055,56 @@ export default class PPNode extends PIXI.Container {
...
   protected async execute(): Promise<boolean> {
     const executedSuccessOld = this.successfullyExecuted;
     let foundChange = false;
     try {
       this.successfullyExecuted = true;
-      if (this.shouldDrawExecution()) {
-        this.renderOutlineThrottled();
+      // if a sockets is a linked and set to manual trigger
+      // ask trigger type if node should execute
+      if (!this.hasLinkedManualTrigger() || this.shouldNodeExecute()) {
```
magnificus, line 1100: we are unfortunately losing the desired encapsulation with this, as now the base node knows about the trigger type and it is affecting the workflow. Also I am concerned about performance here as we force a check of every socket (and print it all too) for every execution.

magnificus (changes requested):
Sorry but I'm a bit concerned about the execution flow now, if you want to feel free to merge, but we need to clean it up a bit

PR #101, opened 2022-05-18
Update behaviour header

fakob:
This PR should visualise the update behaviour for each node in a simple way. Even though this is not perfect, I think it is a good first step.

A small element on the top of every node
* shows the update behaviour
  * if update is selected -> nothing is shown (as this is the default setting)
  * if update is not selected -> a square is shown (sort of like a stop sign)
  * if interval is selected -> the interval frequency is shown
* allows to update now
  * when hovering over a node -> a circle button is shown, clicking it does the same as "update now"

Screen Shot 2022-05-18 at 21 11 39

### DIFF
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -16,6 +16,7 @@ import {
   TSocketType,
 } from '../utils/interfaces';
 import {
+  COLOR_WHITE_TEXT,
   COMMENT_TEXTSTYLE,
   RANDOMMAINCOLOR,
   NODE_TYPE_COLOR,
@@ -29,6 +30,7 @@ import {
   NODE_TEXTSTYLE,
   NODE_WIDTH,
   SOCKET_HEIGHT,
+  SOCKET_WIDTH,
   SOCKET_TYPE,
 } from '../utils/constants';
 import PPGraph from './GraphClass';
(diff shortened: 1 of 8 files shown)


src/classes/NodeClass.ts
```diff
@@ -668,9 +661,7 @@ export default class PPNode extends PIXI.Container {
     return false;
   }
 
-  protected getUpdateBehaviour(): UpdateBehaviour {
-    return new UpdateBehaviour(true, false, 1000);
-  }
+  protected init(): void {}
```
magnificus, line 664: I would highly recommend making the init function private and only call it in PPNode and keep the getUpdateBehaviour in the child classes. This way we can keep the children "pure" and free of side effects.

src/nodes/api/get.tsx
```diff
@@ -12,9 +12,10 @@ const outputContentName = 'Content';
 
 export class Get extends PPNode {
   // default to poll on interval X seconds
-  protected getUpdateBehaviour(): UpdateBehaviour {
-    return new UpdateBehaviour(false, true, 10000);
+  protected init(): void {
```
magnificus, line 15: following up on above comment, I would keep this one as it is and only call the init in the top parent class, to avoid sullying the children in side effects.

magnificus (changes requested):
Cool stuff looking forward to try it, however there is one thing that I think would be better changed, left comments.

src/classes/NodeClass.ts
fakob: I had changed it at the beginning as I did not fully understand it and had problems making it work. Seems that I have now figured it out and made an update. Please check

magnificus (approved):

PR #102, opened 2022-05-22
Removed unused packages and updated packages including react 18

fakob:


fakob:
After this update one might have to
* delete `node_modules`
* run `yarn` again

PR #104, opened 2022-05-28
removed pure node concept and fixed bug

magnificus:


fakob (approved):
Do I understand correctly that you have removed the check for detecting if something has changed? Do you predict much performance impact?

magnificus:
Yes I did, no I don't but it is possible, might re-add a better version of this later, just this optimization was awkward.

PR #106, opened 2022-05-30
Custom function node

magnificus:


src/nodes/data/dataFunctions.tsx
```diff
@@ -421,6 +389,81 @@ export class ForEachLoop extends ForLoop {
...
+      ),
+      new Socket(SOCKET_TYPE.OUT, outDataName, new AnyType()),
+    ];
+  }
+  protected async onExecute(
+    inputObject: any,
+    outputObject: Record<string, unknown>
+  ): Promise<void> {
+    // before every execute, re-evaluate inputs
+    this.adaptInputs(inputObject[anyCodeName]);
```
fakob, line 422: maybe premature, but should we check if the code has changed and only then do the reEvaluation?

fakob (approved):

src/nodes/data/dataFunctions.tsx
magnificus: I think that's fine, then we don't have to store any state, performance impact should be super small, not replacing sockets unless they have changes.

PR #107, opened 2022-05-31
Widget node

fakob:
* Added 3 widget nodes
  * Button
  * Switch
  * Slider
* Added a placeholder node
* Added functionality to not lose links when a node is missing including showing a toast message if this happens

Screenshot 2022-05-31 at 19 24 53
Screenshot 2022-05-30 at 22 34 18


src/classes/GraphClass.ts
```diff
@@ -385,19 +386,50 @@ export default class PPGraph {
     customArgs?: CustomArgs
   ): T {
     // console.log(this._registeredNodeTypes);
-    const nodeConstructor = this._registeredNodeTypes[type]?.constructor;
+    const newArgs: any = {};
+    const placeholderNode = 'Placeholder';
+    let nodeConstructor;
+    let name;
+
+    if (type === placeholderNode) {
```
magnificus, line 394: parsing on string here is a bit awkward IMO, is it instead possible to create a new function?

magnificus: not a massive deal but

magnificus (approved):
nice! added one minor comment

src/classes/GraphClass.ts
magnificus: Wherever you are setting the type to "PlaceHolder" you can instead of doing that call a separate function in graphclass (would be a newly created one) directly which reduces abstraction.

PR #108, opened 2022-06-02
Create widget from socket

fakob:
I added
* the ability to right click on the socket name
  * for inputs, allow connecting widget nodes
  * for outputs allow to connect a label node 
* added a delete socket action for none default sockets
  * I have updated all nodes to use `getDefaultIO` for socket creation

https://user-images.githubusercontent.com/4619772/171721168-19866101-9ec2-4439-8894-c0a2308f10dd.mp4
Screenshot_2022-06-01_at_22 44 44

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -40,7 +40,11 @@ import GraphOverlay from './components/GraphOverlay';
 import ErrorFallback from './components/ErrorFallback';
 import PixiContainer from './PixiContainer';
 import { Image as ImageNode } from './nodes/image/image';
-import { GraphContextMenu, NodeContextMenu } from './components/ContextMenus';
+import {
+  GraphContextMenu,
+  NodeContextMenu,
+  SocketContextMenu,
+} from './components/ContextMenus';
 import { GraphDatabase } from './utils/indexedDB';
 import PPGraph from './classes/GraphClass';
 import {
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -508,21 +514,30 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
+      const contextMenuPosX = Math.min(
+        window.innerWidth - 248,
+        event.data.global.x
+      );
+      const contextMenuPosY = (offset: number) => {
+        return Math.min(window.innerHeight - offset, event.data.global.y);
+      };
       switch (true) {
-        case target instanceof PPSocket:
+        case target.parent instanceof PPSocket && target instanceof PIXI.Text:
```
magnificus, line 527: I know this is old but switching on instanceof is considered ugly, this is when polymorphism pattern would dictate that this is defined inside the classes themselves (so that i dont see those child classes from the outside)

src/classes/GraphClass.ts
```diff
@@ -507,7 +526,42 @@ export default class PPGraph {
     return link;
   }
 
-  checkOldSocketAndUpdateIt<T extends Socket>(
+  addWidgetNode(socket: PPSocket): void {
+    const socketDataType = socket.dataType.constructor.name;
+    const node = socket.getNode();
+    let nodeType: string;
+    switch (socketDataType) {
```
magnificus, line 533: instead of this, add function inside datatype that is overriding, dont switch on type here outside. Similar to the one above.

src/nodes/logViewer.tsx
```diff
@@ -80,4 +72,33 @@ export class LogViewer extends PPNode {
       );
     };
   }
+
+  public getName(): string {
```
magnificus, line 76: good

magnificus (changes requested):
if possible please change the switching on types, otherwise looks good

src/classes/GraphClass.ts
fakob: Good point. I have now submitted a fix for this and also consolidated the addLabelWidget function into this one.

src/App.tsx
magnificus: It would involve making a shared interface (a class called Clickable for example) with a function clickBehaviour (or a better name) that PPSocket, PPNode, and Viewport would extend, then they override those functions, we can look at it, not as important as the one below, feel free to merge before we fix this

magnificus (approved):

PR #109, opened 2022-06-03
Fix issue uploading a graph  not working, when no custom nodes stored

fakob:
I found a bug when trying out the latest graph you sent. Loading did not work because the configure function was expecting a customNodeType object. Your graph did not have that.


magnificus (approved):
this will be solved with the removal of custom node types, but approve for now

PR #110, opened 2022-06-11
prevent pixijs draw crash on undefined

fakob:


fakob:
@magnificus or should I throw an error if it is undefined so it is caught by your try catch and shows the red outline. Might be the better behaviour. What do you think?

magnificus (approved):

magnificus:
sure

fakob:
I updated it to throw an error if width or height is not a finite number. This also covers NaN, which would also end in a webgl crash.

PR #113, opened 2022-06-17
Up down stream node selection

fakob:
* Added option to select
  * all up stream nodes
  * all down stream nodes
  * all nodes of a branch
* updated the update behaviour look

Screenshot 2022-06-17 at 23 23 59


src/App.tsx
```diff
@@ -1471,7 +1471,7 @@ NOTE: save the playground after loading, if you want to make changes to it`
                     <NodeSearchInput
                       {...props}
                       inputRef={nodeSearchInput}
-                      randomMainColor={RANDOMMAINCOLOR}
```
magnificus, line 1474: accidental change? :)

src/classes/NodeClass.ts
```diff
@@ -399,13 +409,30 @@ export default class PPNode extends PIXI.Container {
     }
   }
 
-  getDirectDependents(): { [key: string]: PPNode } {
+  getDirectDependents(
```
magnificus, line 412: I would make a new function for getting parents, and not bloat this one, also, won't dependents of outputsocketarray be a subset of inputsocketarray?

src/classes/NodeClass.ts
```diff
@@ -459,6 +486,45 @@ export default class PPNode extends PIXI.Container {
...
+        combinedDependents[node.id] = node;
+
+        // remove already checked dependents
+        for (const key in newDependents) {
+          if (combinedDependents[key] !== undefined) {
+            delete newDependents[key];
+          }
+        }
+
+        if (Object.values(newDependents).length > 0) {
```
magnificus, line 509: no need to check for this, if its 0 then it wont do anything anyway, so if is not needed

src/classes/NodeClass.ts
```diff
@@ -459,6 +486,45 @@ export default class PPNode extends PIXI.Container {
...
+  ): PPNode[] {
+    const getDirectDependentsAndAccumulateThem = (dependents: {
+      [key: string]: PPNode;
+    }): void => {
+      Object.values(dependents).forEach((node) => {
+        const newDependents: { [key: string]: PPNode } =
+          node.getDirectDependents(true, true, true);
+
+        combinedDependents[node.id] = node;
+
```
magnificus, line 501: instead of playing with this array, just do a filter, something like

getDirectDependentsandAccumulateThem(node.getDirectDependents().filter(dependent => combinedDependents[key] !== undefined))

magnificus: can be separated into two lines as well, but then you wont have to do delete and stateful operations

src/classes/NodeClass.ts
magnificus: so one function for upstream and one for downstream

src/classes/SocketClass.ts
```diff
@@ -270,13 +270,26 @@ export default class Socket extends PIXI.Container {
...
-    // ask the socket whether their children are dependent
-
-    const nodes = this.links.map((link) => link.getTarget().getNode());
-    const filteredNodes = nodes.filter((node) => node.updateBehaviour.update);
-    return filteredNodes;
+  getDirectDependents(
+    ignoreUpdateBehaviour = false,
+    upstream = false
+  ): PPNode[] {
+    if (upstream) {
```
magnificus, line 277: not a fan of all the branches and complexity added to this function, needs to be broken up into two, it was fairly elegant before.

magnificus (changes requested):

src/App.tsx
fakob: The reason I had changed this to all lower case is that react threw an error saying that camelcase naming is only for react props. Not sure why it acts up here as it does not make sense, but at least this change got rid of this error.

src/classes/NodeClass.ts
```diff
@@ -409,6 +419,28 @@ export default class PPNode extends PIXI.Container {
     return currDependents;
   }
 
+  getLinkedNodes(
```
magnificus, line 422: hate to be anal but I dont think this is an improvement, now this clause is copy pasted 3 times, in getDirectDependents and on both clauses below, additionally I think you might end up iterating the same nodes a rapidly scaling number of times, wont "getdepents" upwards also come back to the node firing it ? as it includes downstream?

src/classes/SocketClass.ts
magnificus: this is definitely better now but this whole function be written even tighter and more explicit (less copy paste) as:

return this.links.map(link => (upstream ? link.getSource() : link.getTarget()).getNode())

src/classes/NodeClass.ts
magnificus: I think that you only need to include downstream on the very top node.

magnificus (changes requested):
looking better resolved some comments, but one (two) things lingering

src/classes/NodeClass.ts
```diff
@@ -424,19 +424,22 @@ export default class PPNode extends PIXI.Container {
     includeDownstream = true
   ): { [key: string]: PPNode } {
     const currDependents: { [key: string]: PPNode } = {};
-    if (includeUpstream) {
-      this.inputSocketArray.forEach((socket) => {
-        Object.values(socket.getLinkedNodes(true)).forEach((dependent) => {
+    const goThroughSockets = (
```
magnificus, line 427: this is good, just one more thing youll probably save performance by putting this function outside of this one (in a compiled language it would have been done for you but typescript is not too good at this).

src/classes/UpdateBehaviourClass.ts
```diff
@@ -37,74 +36,73 @@ export default class UpdateBehaviourClass extends PIXI.Container {
...
 
+    this._noUpdateRef.width = 24;
+    this._noUpdateRef.height = 24;
+
     this.redrawAnythingChanging();
   }
 
   redrawAnythingChanging(): void {
     // reset
-    this._updateRef.clear();
```
magnificus, line 70: red, nice

src/classes/NodeClass.ts
magnificus: alright this is starting to look a lot neater

magnificus (approved):
nice work

PR #114, opened 2022-06-19
Reroute node

magnificus:
looks very ugly but needs some other changes to make it look good

### DIFF
--- a/src/nodes/base.ts
+++ b/src/nodes/base.ts
@@ -24,11 +24,8 @@ import { EnumType } from './datatypes/enumType';
 import { BooleanType } from './datatypes/booleanType';
 
 export class Placeholder extends PPNode {
-  constructor(name: string, customArgs: CustomArgs) {
-    super(name, {
-      ...customArgs,
-      color: TRgba.fromString(NODE_TYPE_COLOR.MISSING),
-    });
+  getColor(): TRgba {
+    return TRgba.fromString(NODE_TYPE_COLOR.MISSING);
   }
 
   getCanAddInput(): boolean {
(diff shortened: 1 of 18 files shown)


src/nodes/base.ts
```diff
@@ -321,10 +318,12 @@ export class RangeArray extends PPNode {
...
 export class RandomArray extends PPNode {
+  getColor(): TRgba {
+    return TRgba.fromString(NODE_TYPE_COLOR.INPUT);
+  }
   constructor(name: string, customArgs: CustomArgs) {
     super(name, {
       ...customArgs,
-      color: TRgba.fromString(NODE_TYPE_COLOR.INPUT),
     });
   }
```
fakob, line 328: Could the constructor be removed here as well?

src/nodes/base.ts
```diff
@@ -459,10 +457,12 @@ export class If_Else extends PPNode {
...
 export class Comparison extends PPNode {
+  getColor(): TRgba {
+    return TRgba.fromString(NODE_TYPE_COLOR.TRANSFORM);
+  }
   constructor(name: string, customArgs: CustomArgs) {
     super(name, {
       ...customArgs,
-      color: TRgba.fromString(NODE_TYPE_COLOR.TRANSFORM),
     });
   }
```
fakob, line 468: Could the constructor be removed here as well?

fakob (approved):
Added some comments. Looks good.

src/nodes/base.ts
magnificus: yes indeed it should be

src/nodes/base.ts
magnificus: yes indeed it should be

src/nodes/image/shader.ts
```diff
@@ -3,7 +3,7 @@ import * as PIXI from 'pixi.js';
 import PPGraph from '../../classes/GraphClass';
 import PPNode from '../../classes/NodeClass';
 import UpdateBehaviourClass from '../../classes/UpdateBehaviourClass';
-import { CustomArgs, TRgba } from '../../utils/interfaces';
+import { CustomArgs, TRgba} from '../../utils/interfaces';
```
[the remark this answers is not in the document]
magnificus: dont know might be autocorrected (?)

PR #118, opened 2022-06-30
fix regression with label and table node

fakob:


magnificus (approved):

src/nodes/table.tsx
```diff
@@ -74,7 +74,7 @@ export class Table extends PPNode {
   }
 
   getOpacity(): number {
```
magnificus, line 76: I would add a comment to explain why this is needed

PR #120, opened 2022-07-13
Switching column and row order and execution visualisation

fakob:
* Added support for switching column and row order for multiply node
* Added support for turning of execution visualisation

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -133,6 +133,8 @@ const App = (): JSX.Element => {
   const [isCurrentGraphLoaded, setIsCurrentGraphLoaded] = useState(false);
   const [actionObject, setActionObject] = useState(null); // id and name of graph to edit/delete
   const [showComments, setShowComments] = useState(false);
+  const [showExecutionVisualisation, setShowExecutionVisualisation] =
+    useState(true);
   const [remoteGraphs, setRemoteGraphs, remoteGraphsRef] = useStateRef([]);
   const [graphSearchItems, setGraphSearchItems] = useState<
     IGraphSearch[] | null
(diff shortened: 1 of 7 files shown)


src/nodes/draw/draw.tsx
```diff
@@ -520,18 +523,24 @@ export class DRAW_Multiplier extends DRAW_Base {
...
     let numPlaced = 0;
-    for (let x = 0; x < numX; x++) {
-      for (let y = 0; y < numY && numPlaced < total; y++, numPlaced++) {
-        const currentIndex = x + inputObject[totalNumberName] * y;
+
+    for (let i = 0; i < numI; i++) {
+      for (let j = 0; j < numJ && numPlaced < total; j++, numPlaced++) {
+        const currentIndex = numPlaced;
+
+        if (changeDrawingOrder) {
```
magnificus, line 535: instead of changing them and changing them back (stateful operation), just declare new variables:

x = changeDrawingOrder ? j : i;

etc.

src/App.tsx
```diff
@@ -133,6 +133,8 @@ const App = (): JSX.Element => {
   const [isCurrentGraphLoaded, setIsCurrentGraphLoaded] = useState(false);
   const [actionObject, setActionObject] = useState(null); // id and name of graph to edit/delete
   const [showComments, setShowComments] = useState(false);
+  const [showExecutionVisualisation, setShowExecutionVisualisation] =
```
magnificus, line 136: This is OK but we are starting to suffer some performance concerns from this massive top component, can this state be moved down? Does anything up here use this?

magnificus (changes requested):

src/nodes/draw/draw.tsx
fakob: Ah of course. Good idea.

fakob:
I have also added storing showExecutionVisualisation boolean in the graph.

magnificus (approved):

PR #124, opened 2022-07-24
Added check for localhost

fakob:
Otherwise we get a CORS error when running locally. Or is there a better way of doing this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@uiw/react-codemirror": "^4.3.2",
     "color": "^4.0.1",
     "core-js": "^3.22.7",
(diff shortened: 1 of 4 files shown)


magnificus (approved):
we should be able to ignore CORS restrictions as we are hosting, what is inside this buldinfo?

magnificus:
if its the electron process then it can be fixed by setting up launch options

fakob:
I have now added cors middleware and enabled cors, according to the expressjs guidelines - https://expressjs.com/en/resources/middleware/cors.html

magnificus (approved):

PR #129, opened 2022-08-08
App refactor

magnificus:


fakob (approved):

magnificus:
not happy with everything here, might close or will update, will not merge as is

PR #137, opened 2022-09-04
Connect to matching socket 

fakob:


src/utils/utils.ts
```diff
@@ -557,3 +558,24 @@ export const isVariable = (
...
   }
 };
+
+export const getMatchingSocketIndex = (
+  socket: PPSocket,
+  socketArray: PPSocket[]
+): number => {
+  const socketDataTypeName = socket.dataType.getName();
+
+  const indexExactMatch = socketArray.findIndex((socketInArray) => {
```
magnificus, line 568: See if possible to not match string but instead look at class

magnificus (approved):

PR #147, opened 2022-09-22
Drag link on node

fakob:
This feature allows you to drag connections from inputs or outputs directly onto another node.  In addition I have also added both preferred input and output socket index. The logic is:
* if node has a preferred socket index without connection > use it
* if node has a socket with the same type without connection > use it
* if node has a AnyType socket without connection > use it
* if node has a socket without connection > use it
* if all sockets have connections, connect to preferred socket, else connect to first socket

I have gone through some of the nodes and added preferred sockets were I thought it fits.

### DIFF
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -38,6 +38,7 @@ import PPGraph from './GraphClass';
 import Socket from './SocketClass';
 import {
   calculateAspectRatioFit,
+  getMatchingSocketIndex,
   getNodeCommentPosX,
   getNodeCommentPosY,
 } from '../utils/utils';
@@ -1177,8 +1178,8 @@ ${Math.round(this._bounds.minX)}, ${Math.round(
     this.onMoveHandler = this._onPointerMove.bind(this);
 
     this.on('pointerdown', this._onPointerDown.bind(this));
-    this.on('pointerup', this._onPointerUpAndUpOutside.bind(this));
-    this.on('pointerupoutside', this._onPointerUpAndUpOutside.bind(this));
+    this.on('pointerup', this._onPointerUp.bind(this));
+    this.on('pointerupoutside', this._onPointerUpOutside.bind(this));
     this.on('pointerover', this._onPointerOver.bind(this));
     this.on('pointerout', this._onPointerOut.bind(this));
     this.on('dblclick', this._onDoubleClick.bind(this));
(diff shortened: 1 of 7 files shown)


src/classes/NodeClass.ts
```diff
@@ -293,6 +294,14 @@ export default class PPNode extends PIXI.Container {
     return new AnyType();
   }
 
+  getPreferredInputSocketIndex(): number | undefined {
```
magnificus, line 297: Can we instead of using indices here use string for name? I have tried to remove all references to indices in the nodes in order to make adding/removing inputs/outputs not cause issues.

src/nodes/table.tsx
```diff
@@ -40,6 +40,15 @@ export class Table extends PPNode {
   protected getActivateByDoubleClick(): boolean {
     return true;
   }
+
+  getPreferredInputSocketIndex(): number {
```
magnificus, line 44: 1 here means nothing to me so a string will be much more clear I think

src/utils/utils.ts
```diff
@@ -560,19 +560,54 @@ export const isVariable = (
...
-  socketArray: PPSocket[]
-): number => {
-  const indexExactMatch = socketArray.findIndex((socketInArray) => {
-    return socketInArray.dataType.constructor === socket.dataType.constructor;
-  });
+  node: PPNode,
+  isOutput = false
+): number | undefined => {
+  const socketArray = isOutput ? node.outputSocketArray : node.inputSocketArray;
+  if (socketArray.length > 0) {
```
magnificus, line 567: no need to check for size 0, you either find it or you don't, no matter the size

src/utils/utils.ts
```diff
@@ -560,19 +560,54 @@ export const isVariable = (
...
+    // if the node has a preferredInputSocket (with no connection), use it
```
magnificus, line 568: This function can be made a lot shorter, I would do something like this:

targetArray = isOutput ? node.getPreferredOutputSocket : node.getPreferredInputSocket();
mainTarget= targetArray.find("that node existsand has no link to it")
anyTarget= targetArray.find("anytype that has no link to it")
lastChoice = targetArray.find("any available socket")
return mainTarget || anyTarget || lastChoice.

magnificus (changes requested):

src/classes/NodeClass.ts
magnificus: Also, lets not make it possible to return undefined, lets just make the default some silly (but explicit) string like
"MyPreferredInputSocket"

src/utils/utils.ts
magnificus: bonus, making this a oneliner will save performance as the last two don't have to be evaluated if something is found in first (though lowers readability so probably not worth it)

src/utils/utils.ts
fakob: There is a slight difference.
* If you try to connect to a node which does not have an input/output, you should not be able to -> undefined
* If you try to connect to a node which has an input/output, you should be connected regardless of if it is a perfect match or just whatever -> index

src/classes/NodeClass.ts
fakob: Good idea.

src/nodes/table.tsx
fakob: You are right

src/utils/utils.ts
fakob: I have made it into a "oneliner", but broken it down into functions. This should cover both the readability and the performance aspect of it.

magnificus (approved):

PR #149, opened 2022-09-24
basic functionality and some actions undo/redoable

magnificus:


src/App.tsx
```diff
@@ -570,49 +571,47 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
       const modKey = isMac ? e.metaKey : e.ctrlKey;
       // console.log(e.key);
       if (!isEventComingFromWithinTextInput(e)) {
+        e.preventDefault();
```
fakob, line 574: I had added `preventDefault()` on the case level is so it does not preventDefault if no case match. The believe was that I did not want to preventDefault if a shortcut is used by another element. Not sure what exactly that use case would be, but could you check if these use cases still work? 
* shortcuts like ctrl+a to select all in nodes like codeEditor and textEditor
* browser shortcuts

src/App.tsx
```diff
@@ -570,49 +571,47 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
               setIsGraphSearchOpen((prevState) => !prevState);
               break;
             case 'e':
-              e.preventDefault();
               setShowEdit((prevState) => !prevState);
               break;
+            case 'z':
+              ActionHandler.undo();
+              break;
+            case 'r':
```
fakob, line 595: I suggest to use modKey && e.shiftKey && 'z' for redo. This is the browser convention.
And modKey && 'r' is usually used for refresh. I would probably not touch that shortcut.

fakob: Ok, just saw that in Windows world the convention in the browser is mixed between
* ctrl+Y (most apps, including Office)
* ctrl+shift+Z (Miro)
I usually check what Miro is doing as it is similar and widespread both on windows and mac. Then we do not have to differentiate between platforms.

fakob (approved):

src/App.tsx
magnificus: hmm OK can check

src/App.tsx
magnificus: OK

magnificus: switching to ctrl shift z

src/App.tsx
fakob: The browser command reloading the tab stopped working now (ctrl + r). I suggest to revert this change and add preventDefault again for every single shortcut.

fakob: Copy pasting also stopped working.

magnificus: Alright thought I had but apparently I forgot

PR #150, opened 2022-09-27
Move trigger sockets out of socket array

fakob:
* I have introduced a nodeTriggerSocketArray, which is being used if you right click add a node trigger socket
  * Though it is still possible to use triggerType on normal inputs
  * The node trigger sockets are displayed on the top before the outputs
* I have updated the look of trigger types to be square
* I have updated the addTriggerInput function to also turn off "Update on change" as this is probably always what you want when adding a nodeTriggerSocket

Screen Shot 2022-09-27 at 21 52 10

### DIFF
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -282,14 +282,10 @@ export default class PPNode extends PIXI.Container {
     return this.constructor.toString();
   }
 
-  addSocket(socket: Socket, isNodeTrigger = false): void {
+  addSocket(socket: Socket): void {
     const socketRef = this.addChild(socket);
     switch (socket.socketType) {
       case SOCKET_TYPE.IN: {
-        if (isNodeTrigger) {
-          this.nodeTriggerSocketArray.push(socketRef);
-          return;
-        }
         this.inputSocketArray.push(socketRef);
         break;
       }
@@ -298,6 +294,12 @@ export default class PPNode extends PIXI.Container {
         break;
       }
     }
+  }
+
+  addTriggerSocket(socket: Socket): void {
+    const socketRef = this.addChild(socket);
+    this.nodeTriggerSocketArray.push(socketRef);
+    return;
   }
 
   getDefaultType(): AbstractType {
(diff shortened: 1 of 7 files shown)


src/classes/NodeClass.ts
```diff
@@ -275,10 +282,14 @@ export default class PPNode extends PIXI.Container {
     return this.constructor.toString();
   }
 
-  addSocket(socket: Socket): void {
```
magnificus, line 278: can we make a new function for this instead? so we dont have to pass in these conditionals to function

src/classes/NodeClass.ts
```diff
@@ -375,9 +384,7 @@ export default class PPNode extends PIXI.Container {
         this.resizedNode();
       }
 
-      const sockets = nodeConfig.socketArray;
-
-      sockets.forEach((item: SerializedSocket) => {
+      const mapSocket = (item: SerializedSocket, isNodeTrigger = false) => {
```
magnificus, line 387: again lets not pass in this conditional to all of our functions if we can avoid it

src/classes/NodeClass.ts
```diff
@@ -681,8 +696,15 @@ export default class PPNode extends PIXI.Container {
     this.resizeNode(this.minNodeWidth, this.calculatedMinNodeHeight);
   }
 
-  getAllSockets(): Socket[] {
-    return this.inputSocketArray.concat(this.outputSocketArray);
+  getAllInputSockets(): Socket[] {
+    return this.inputSocketArray.concat(this.nodeTriggerSocketArray);
+  }
+
+  getAllSockets(includeNodeTriggerSocketArray = true): Socket[] {
```
magnificus, line 703: Would recommend making a new function, maybe rename the old one to getDataSockets()

and then getAllSockets() returns getDataSockets().concat(this.nodeTriggerSocketArray

magnificus (changes requested):
Nice, looking forward to trying it, added some comments about how we can reduce conditionals in the code

src/classes/NodeClass.ts
fakob: sounds good

src/classes/NodeClass.ts
fakob: I suggest to leave the top-level conditional for the mapSocket. This function is only local. It was introduced to not have to duplicate code, which I would have to do if I make 2 unique ones.

src/classes/NodeClass.ts
fakob: Good idea!

src/classes/NodeClass.ts
magnificus: OK it is only local i see, then ait

magnificus (approved):

PR #152, opened 2022-10-04
Text node updates and fixes

fakob:
* Added option which creates a rich text node if you paste html or plain text right into the canvas
  * Updated the way we are copying nodes -> now using the text/html mime type to better differentiate what is copy/pasted
* Added autoHeight option for the rich text node (on by default)
* Made the min-size of the rich text node smaller
* Moved the rest of the hybrid node functions into the hybridNode class

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -391,7 +391,7 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
             e.preventDefault();
             if (currentGraph.current.selection.selectedNodes.length < 1) {
               const mouseWorld = viewport.current.toWorld(mousePosition);
-              currentGraph.current.createAndAddNode('TextEditor', {
+              currentGraph.current.addNewNode('TextEditor', {
                 nodePosX: mouseWorld.x,
                 nodePosY: mouseWorld.y,
                 initialData: { html: text },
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -374,19 +373,63 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
-      try {
-        const json = JSON.parse(textFromClipboard) as SerializedSelection;
-        if (json.version) {
-          e.preventDefault();
-          const pastedNodes = await currentGraph.current.pasteNodes(json, true);
-          console.log(pastedNodes);
-        } else {
-          console.error('Clipboard does not contain valid node data');
+      if (!isEventComingFromWithinTextInput(e)) {
+        const clipboardBlobs = await getDataFromClipboard();
```
magnificus, line 377: these two statements share a ton of code and should be consolidated into common function avoid copy pasted code

fakob: You are right, it looks like there is a lot of copy pasted code, but these 4 options are all slightly different. I have now created a common function, but it is not as small as you would have hoped I guess.

src/nodes/textEditor/textEditor.tsx
```diff
@@ -245,6 +291,59 @@ export class TextEditor extends HybridNode {
...
+          const editorHeight = Math.ceil(
+            document.getElementById(this.id).getBoundingClientRect().height /
+              PPGraph.currentGraph.viewport.scale.x
+          );
+          this.resizeNode(this.nodeWidth, editorHeight);
+        }
+      };
+
+      // workaround to get ref of editor to be used as mounted/ready check
+      useEffect(() => {
```
magnificus, line 305: looks a lil bit sus

magnificus (approved):

src/nodes/textEditor/textEditor.tsx
fakob: This is what slatejs people have suggested as there is nothing native built in (yet).
https://github.com/ianstormtaylor/slate/issues/3618#issuecomment-653635644

PR #153, opened 2022-10-09
Switch to glide data grid 2

fakob:
* Switched to https://github.com/glideapps/glide-data-grid
* Removed CSV output option
* Changed preferredCondition to also work if the socket is hidden
* Fixed an issue with getMatchingSocket function

Known issues:
* Currently it is not possible to manually add rows or columns
* Pasting data into the grid, does not support adding more rows or columns in case it is needed


package.json
```diff
@@ -40,7 +40,9 @@
     "javascript-time-ago": "^2.3.10",
     "json5": "^2.2.0",
     "jsonpath-plus": "^7.2.0",
+    "lodash": "^4.17.21",
```
magnificus, line 43: do we need both lodash and lodash contrib? could we lose one of them?

src/utils/utils.ts
```diff
@@ -138,7 +138,10 @@ export const limitRange = (
   upperLimit: number
 ): number => {
   // value || 0 makes sure that NaN s are turned into a number to work with
-  return Math.min(Math.max(value || 0, lowerLimit || 0), upperLimit || 0);
+  if (lowerLimit < upperLimit) {
```
magnificus, line 141: can be rewritten to lose the branch, would make the function like this

min = Math.min(lowerLimit, upperLimit)
max = Math.max(lowerLimit,upperLimit)

return Math.min(Math.max(min,value), value)

Better to not check for null and crash instead of something tries to do something so illegal

src/utils/utils.ts
```diff
@@ -620,3 +627,24 @@ export const getMatchingSocket = (socket: PPSocket, node: PPNode): PPSocket => {
...
+  newSize: number,
+  defaultValue: any
+): any[] => {
+  return [
+    ...arr,
+    ...Array(Math.max(newSize - arr.length, 0)).fill(defaultValue),
+  ];
+};
+
+export const getLongestArrayInArray = (arrayOfArrays): number => {
```
magnificus, line 642: return 0 instead of undefined, and there is no need to branch here, just give the reduce a default value, this function can probably be just one line

magnificus (changes requested):

src/nodes/table.tsx
```diff
@@ -156,98 +161,190 @@ export class Table extends HybridNode {
...
-          .loadData(props.dataArray)
-          .change(handleOnChange);
-        this.xSpreadSheet.on('cells-selected', handleOnSelect);
+        loadSheet();
       }, []);
 
       useEffect(() => {
-        this.xSpreadSheet.loadData(props.dataArray);
-      }, [props.dataArray]);
+        if (props.doubleClicked) {
```
magnificus, line 201: should this be in the hybrid node base class?

package.json
fakob: lodash is a peer dependency of the data grid. lodash contrib is as far as I know, an additional package. They are not the same.

src/utils/utils.ts
fakob: Nice. will do

package.json
magnificus: hmm if its a peer dependency wouldnt it be just in the lock file? I might be wrong

magnificus (approved):

PR #154, opened 2022-10-11
Html renderer node

fakob:
Added an html render node which allows to embed any iframe code (youtube, vimeo, website). Unfortunately we can not embed our github page, as it disallows embedding.

Screen Shot 2022-10-11 at 22 17 11

### DIFF
--- a/src/nodes/draw/html.tsx
+++ b/src/nodes/draw/html.tsx
@@ -1,0 +1,170 @@
+import React, { useEffect, useState } from 'react';
+import { Box, ThemeProvider } from '@mui/material';
+import { ErrorBoundary } from 'react-error-boundary';
+import ErrorFallback from '../../components/ErrorFallback';
+import PPSocket from '../../classes/SocketClass';
+import PPGraph from '../../classes/GraphClass';
+import PPNode from '../../classes/NodeClass';
+import { CodeType } from '../datatypes/codeType';
+
+import { convertToString } from '../../utils/utils';
+import { CustomArgs } from '../../utils/interfaces';
+import { SOCKET_TYPE, customTheme } from '../../utils/constants';
+
+const outputSocketName = 'output';
+const inputSocketName = 'input';
+
+export class HtmlRenderer extends PPNode {
+  update: (newHeight?) => void;
+
+  protected getIsHybrid(): boolean {
+    return true;
+  }
+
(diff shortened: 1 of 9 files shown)


src/nodes/draw/html.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { CustomArgs } from '../../utils/interfaces';
+import { SOCKET_TYPE, customTheme } from '../../utils/constants';
+import HybridNode from '../../classes/HybridNode';
+
+const inputSocketName = 'Html';
+
+export class HtmlRenderer extends HybridNode {
+  update: (newHeight?) => void;
+
+  protected getIsHybrid(): boolean {
```
magnificus, line 20: this function should not be needed anymore I think

src/nodes/draw/html.tsx
```diff
@@ -0,0 +1,170 @@
...
+      nodeHeight,
+      minNodeWidth: nodeWidth / 2,
+      minNodeHeight: nodeHeight / 2,
+    });
+
+    if (customArgs?.initialData) {
+      this.setInputData(inputSocketName, customArgs?.initialData);
+    }
+
+    // when the Node is added, create the container and react component
```
magnificus, line 82: could we have this in the common hybrid node class? It seems to be roughly the same for all hybrid nodes

magnificus (approved):

src/nodes/draw/html.tsx
fakob: Ah, yes. Good catch.

fakob: Updated

src/nodes/draw/html.tsx
fakob: To be honest, I really do not know how to do this. I have to pass in the `ParentComponent` and `props` which are custom for every hybrid node. And this has to happen after the node has been created.
Do you want to give it a shot?

magnificus: Then I would make a function in hybridnode to be overridden which returns this parentcomponent and props, I can take a look if you want

fakob: Yes, please do. Do you want to push to this branch or should I merge and you create a separate PR?

magnificus: I can do it on this as long as i dont forget

PR #155, opened 2022-10-14
Add option to replace a node

fakob:
Added an option to replace a node by
* right clicking it and choosing replace with in the context menu
* selecting a node and click ctrl/cmd + f

The action is undoable. Woohoo :-)

https://user-images.githubusercontent.com/4619772/195927611-a4051ee2-f113-4d1d-8e5f-933ee939c84b.mp4

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1086,6 +1086,7 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
           PPGraph.currentGraph.nodes[oldId],
           newNode
         );
+        newNode.executeOptimizedChain();
         PPGraph.currentGraph.selection.selectNodes([newNode]);
         PPGraph.currentGraph.selection.drawRectanglesFromSelection();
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1056,46 +1056,97 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
+      };
 
-      setNodeSearchActiveItem(selected);
-      setIsNodeSearchVisible(false);
-    };
-    const undoAction = async () => {
-      PPGraph.currentGraph.removeNode(PPGraph.currentGraph.nodes[referenceID]);
-    };
-    ActionHandler.performAction(action, undoAction);
+      const action = async () => {
```
magnificus, line 1095: looks good, but could this be in another file? just to avoid bloating App.tsx which is quickly becoming a god file :)

magnificus (approved):

src/App.tsx
fakob: Done

PR #156, opened 2022-10-16
Improve node search

fakob:
This is the first part of improving the node search.
* Improved the node search. It has now a more intuitive weighting and prioritises the name over the description
* Update the node search styling to show the matched parts and the match count
* Moved some functions out of App.tsx

Next step could be to go improve node names and descriptions.

Screenshot 2022-10-16 at 13 43 00


src/App.tsx
```diff
@@ -1559,18 +1372,28 @@ NOTE: save the playground after loading, if you want to make changes to it`
...
                 isOptionEqualToValue={(option, value) =>
-                  option.title === value.title
+                  option.name === value.name
                 }
                 // open
                 PopperComponent={(props) => <GraphSearchPopper {...props} />}
                 value={graphSearchActiveItem}
                 getOptionDisabled={(option) => option.isDisabled}
-                getOptionLabel={(option) => option.name}
+                getOptionLabel={(option) => (option as any).name}
```
magnificus, line 1381: is as any needed?

src/components/Search.tsx
```diff
@@ -79,3 +228,126 @@ export const NodeSearchInput = (props) => {
...
+        key: title,
+        description: obj.description,
+        hasInputs: obj.hasInputs.toString(),
+      };
+    })
+    .sort(
+      (a, b) => a.name.localeCompare(b.name, 'en', { sensitivity: 'base' }) // case insensitive sorting
+    )
+    .filter((node) =>
+      addLink ? node.hasInputs === 'true' : 'true'
```
magnificus, line 248: looks a bit odd that the boolean is a string

magnificus (approved):

src/components/Search.tsx
fakob: Agree. Will fix.

src/App.tsx
fakob: There was a type error which was not caught before, but now it was. Not sure why. According to the MUI documentation:
> If [Autocomplete] used in free solo mode, it must accept both the type of the options and a string.

I have removed the any now and am checking for type.

PR #160, opened 2022-10-20
Jump to node

fakob:
Added a widget to jump to another node

https://user-images.githubusercontent.com/4619772/197048740-a76910b8-d35e-435b-b6e2-3500b3360f75.mp4


src/nodes/datatypes/enumType.tsx
```diff
@@ -9,19 +9,31 @@ export type EnumStructure = { text: string; value?: any }[];
...
     this.options = inOptions;
     this.onChange = onChange;
+    this.setOptions = setOptions;
   }
 
   getName(): string {
     return 'Enum';
   }
 
+  setSetOptions(newSetOptions: () => EnumStructure): void {
```
magnificus, line 29: Hmm this is stateful, can we instead provide this on creation? Or maybe just make it deferred entirely? So that its for example defined by a callback to get the latest options instead?

src/nodes/utility/utility.ts
```diff
@@ -63,3 +67,69 @@ export class Reroute extends PPNode {
...
+      ...customArgs,
+    });
+
+    this.onConfigure = (): void => {
+      const nodeNameDropdown = this.getSocketByName(selectNodeName)
+        .dataType as EnumType;
+      // set the options on Configure
+      nodeNameDropdown.options = this.nodeArrayOptions();
+      // pass in a callback to renew options when dropdown is openend
+      nodeNameDropdown.setSetOptions(this.nodeArrayOptions);
```
magnificus, line 103: same as above, this is stateful, callback is good idea, but lets commit fully to it instead and never set any data at all

magnificus (approved):
Looks good apart from the stateful enum stuff, would be great to change

src/nodes/utility/utility.ts
fakob: Ah good catch. Setting the options is not needed. I am instead calling the setOptions function also when mounting the component.

src/nodes/datatypes/enumType.tsx
fakob: I am sorry, could you give me an example. Not sure what you had in mind :-)

PR #162, opened 2022-10-22
state nodes for array and number

magnificus:


src/nodes/state/stateNodes.ts
```diff
@@ -1,32 +1,105 @@
...
+  protected abstract add(): void;
+  protected abstract remove(): void;
+}
+export class ArrayState extends StateNode {
+  protected add(): void {
+    const state: any[] = this.getInputData('State');
     state.push(this.getInputData('Input'));
-    this.setInputData('Input', state);
+    this.setInputData('State', state);
+    this.executeOptimizedChain();
```
fakob, line 66: Not sure why, but when I had a codeEditor node hooked up to the output of the ArrayState, it did not show the values in the codeEditor. Does this work for you?

fakob (approved):

src/nodes/state/stateNodes.ts
magnificus: let me check

magnificus: when i use the github web thing i can use nodes on the other side, though i noticed clear is broken will fix

PR #163, opened 2022-10-22
Fix render issue of hybrid nodes

fakob:


src/classes/NodeClass.ts
```diff
@@ -124,6 +125,14 @@ export default class PPNode extends PIXI.Container {
     return minHeight;
   }
 
+  public getDefaultNodeWidth(): number {
+    return;
```
magnificus, line 129: these fellas look broken, should they be removed?

magnificus (approved):
Nice but lets make defaultnodewith default not return undefined

src/classes/NodeClass.ts
fakob: How would I allow for setting a defaultNodeWidth, but if none is set you would fall back to the minNodeWidth which I was doing like this:
```
this.nodeWidth = this.getDefaultNodeWidth() ?? this.getMinNodeWidth();
```
Is there another way of doing this?

magnificus: what about just making this.getDefaultNodeWidth() return this.getMinNodeWidth() by default? :), then you can just do

this.nodeWidth = this.getDefaultNodeWidth()

fakob: Hmm, I had thought about it, but then for what ever reason came to the conclusion that this would not work. Well, it seems to work :-)
Thanks for pointing it out.

PR #165, opened 2022-10-23
Fixing editor getting non string values

fakob:


fakob:
@magnificus or should this responsibility rather move inside of the <CodeEditor> component?

magnificus (approved):

magnificus:
yeah i guess that could be even cleaner indeed

PR #168, opened 2022-10-29
node-updates-2

fakob:
* When right clicking a socket or its name, you can connect a node. Now, if it is an input socket, the data is transferred over to the new node. This way you can break out data into separate nodes. This also works when there was a previous connection.
  * Before the right clicking was only possible on the text. As hybrid nodes do not have a text, I have removed the "right click a socket removes the link" and moved it into the context menu. This might be controversial, but I was not sure how else to do this
  * Added more defaultInput/OutputWidget definitions for some types
  * Added option for Label, Constant node to receive initial data
* Added optional date string input for DateAndTime node
* Added plain text output for TextEditor
* Fixed txt import
* Fixed socket right click no longer working due to hierarchy change in the socket

https://user-images.githubusercontent.com/4619772/198852007-b572716f-1d64-4efd-bda7-7f30414cd748.mp4

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -176,8 +176,10 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
     console.log(acceptedFiles, fileRejections);
 
     const dropPoint = viewport.current.toWorld(
-      new PIXI.Point(event.clientX, event.clientY)
+      // no event exists in case graph gets loaded from file
+      new PIXI.Point(event?.clientX ?? 0, event?.clientY ?? 0)
     );
+
     let nodePosX = dropPoint.x;
     const nodePosY = dropPoint.y;
     const newNodeSelection: PPNode[] = [];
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -176,8 +176,10 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
     console.log(acceptedFiles, fileRejections);
 
     const dropPoint = viewport.current.toWorld(
-      new PIXI.Point(event.clientX, event.clientY)
+      // no event exists in case graph gets loaded from file
+      new PIXI.Point(event?.clientX ?? 0, event?.clientY ?? 0)
```
magnificus, line 180: possible to not null check? event should never be null right?

magnificus (approved):

src/App.tsx
fakob: We are using the open file dialog function included in react-dropzone which does not return an event. https://react-dropzone.js.org/#!/Opening%20File%20Dialog%20Programmatically
I guess we could write our own, but it is convenient to not do that and use theirs.

PR #170, opened 2022-10-30
simplified naming, fixed nodes showing their actual names in graph as…

magnificus:
… opposed to classname

### DIFF
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -51,7 +51,8 @@ export default class PPNode extends PIXI.Container {
   isHovering: boolean;
 
   id: string;
-  type: string; // Type
+  type: string;
+  overrideName = '';
   nodePosX: number;
   nodePosY: number;
   nodeWidth: number;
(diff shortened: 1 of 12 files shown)


src/classes/NodeClass.ts
```diff
@@ -102,9 +103,18 @@ export default class PPNode extends PIXI.Container {
...
 
-  // we should migrate all nodes to use these functions instead of specifying the field themselves in constructor
+  public getDisplayName(): string {
+    return this.overrideName !== ''
+      ? this.overrideName + '\t(' + this.getName() + ')'
+      : this.getName();
+  }
+
   public getName(): string {
-    return this.name;
```
fakob, line 107: There still seem to be some references to name around, which need updating. For example in here - `InspectorContainer.tsx` and `FloatingNodeMenu.tsx`.

Screenshot 2022-10-30 at 21 27 57
Screenshot 2022-10-30 at 21 28 16

fakob (approved):

src/classes/NodeClass.ts
magnificus: ah, ill update

magnificus:
I might decline this, not sure its worth the hassle, TBD

PR #171, opened 2022-11-07
Updated table node

fakob:
* Added option to move columns
* Added option to resize columns
  * size of the column is not stored though
* Added column menu to add columns to left/right and delete column
* Added add column button at the right
* Added add row button at the bottom

Screen Shot 2022-11-07 at 21 59 20


magnificus (approved):
Cool, this one is quite complex but because its encapsulated within this node its all good, looks promising

PR #172, opened 2022-11-10
Selection removed from node, undo/redo for move finally

magnificus:


fakob (approved):
Very nice. Great work!

PR #174, opened 2022-11-13
Fix trigger socket not being removed and resize issue

fakob:
Fixes
* trigger sockets not actually being removed
* the node not resizing, when removing sockets in general
* trigger connections not being copied
* replacing nodes with connected triggers

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -886,26 +886,21 @@ export default class PPGraph {
   }
 
   reconnectLinksToNewNode(oldNode: PPNode, newNode: PPNode): void {
-    //check nodeTriggers
-    for (let i = 0; i < oldNode.nodeTriggerSocketArray.length; i++) {
-      const oldInputSocket = oldNode.nodeTriggerSocketArray[i];
-      const newInputSocket = newNode.nodeTriggerSocketArray[i];
-      this.checkOldSocketAndUpdateIt(oldInputSocket, newInputSocket, true);
-    }
+    const checkArrayForOldSocketAndUpdateIt = (
+      nameOfArrayToCheck: string,
+      isInput = true
+    ): void => {
+      for (let i = 0; i < oldNode[nameOfArrayToCheck].length; i++) {
+        const oldInputSocket = oldNode[nameOfArrayToCheck][i];
+        const newInputSocket = newNode[nameOfArrayToCheck][i];
+        this.checkOldSocketAndUpdateIt(oldInputSocket, newInputSocket, isInput);
+      }
+    };
 
-    //check inputs
-    for (let i = 0; i < oldNode.inputSocketArray.length; i++) {
-      const oldInputSocket = oldNode.inputSocketArray[i];
-      const newInputSocket = newNode.inputSocketArray[i];
-      this.checkOldSocketAndUpdateIt(oldInputSocket, newInputSocket, true);
-    }
-
-    //check outputs
-    for (let i = 0; i < oldNode.outputSocketArray.length; i++) {
-      const oldOutputSocket = oldNode.outputSocketArray[i];
-      const newOutputSocket = newNode.outputSocketArray[i];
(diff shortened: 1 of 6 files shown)


src/classes/GraphClass.ts
```diff
@@ -882,14 +882,21 @@ export default class PPGraph {
   }
 
   reconnectLinksToNewNode(oldNode: PPNode, newNode: PPNode): void {
-    //disconnect inputs
+    //check nodeTriggers
```
magnificus, line 885: these three blocks are super similar (copy pasted), could we make a common function that they all use?

src/classes/SocketClass.ts
```diff
@@ -308,6 +308,11 @@ export default class Socket extends PIXI.Container {
 
   destroy(): void {
     this.removeLink();
+    this.getNode().nodeTriggerSocketArray =
```
magnificus, line 311: what about just concating nodetriggersocketarray with the normal socket array here? to avoid copy pasting

magnificus (changes requested):
added two comments, copy paste concerns

src/classes/GraphClass.ts
fakob: Makes sense. Will do

src/classes/SocketClass.ts
fakob: The way this was written was that new nodeTrigger and input socket arrays are written, filtering out the destroyed one. This has to be done separate. 
I guess I could write a separate function which finds the socket and returns the array and then call it for
* nodeTriggerSocketArray
* inputSocketArray
* outputSocketArray
This would avoid the copy pasting at least.

src/classes/GraphClass.ts
fakob: Updated, though I did not know of another way then using the name strings of the array. @magnificus is that ok, or is there a better way to do this?

src/classes/SocketClass.ts
fakob: Same here. Please check

src/classes/GraphClass.ts
magnificus: What about this?

function checkAndUpdateSocketArray(oldArray, newArray) : void {
   oldArray.forEach((socket,index) => this.checkOldSocketAndUpdateIt(socket,newArray[index])
}

checkAndUpdateSocketArray(oldNode.nodeTriggerSocketArray, newNode.nodeTriggerSocketArray)
checkAndUpdateSocketArray(oldNode.inputSocketArray, newNode.inputSocketArray)
checkAndUpdateSocketArray(oldNode.outputSocketArray, newNode.outputSocketArray)

this can be further reduced but just this is much better IMO

for more minimal (but maybe harder to understand):

function checkAndUpdateSocketArray(oldNode, newNode, field) : void {
   oldArray[field].forEach((socket,index) => this.checkOldSocketAndUpdateIt(socket,newNode[field][index])
}

["nodeTriggerSocketArray", "inputSocketArray", "outputSocketArray"].forEach(field => checkAndUpdateSocketArray(oldNode,newNode,field);


etc.

src/classes/SocketClass.ts
magnificus: strategies above should work here too I think

fakob: I tried, but it did not work. The array is filtered and its variable is then overwritten. When I write a function which takes this array variable, it seems that I can not overwrite it within the function.

src/classes/GraphClass.ts
fakob: done

src/classes/SocketClass.ts
magnificus: the field thing should work here, but at the very least, lets make 
"socket.name === this.name && socket.socketType === this.socketType"

into a common function

magnificus: This would probably be better on the node as a function

RemoveSocket(socketin)

fakob: Moved it onto the node level

magnificus (approved):

PR #179, opened 2022-11-16
New macros

magnificus:


fakob (approved):
Nice work, I think this layout makes a lot of sense.

Some additional input after testing in codesandbox:
* I had tried with 2 or more parameter inputs in the macro, but the instance always only showed one
* it would be great if there was a dropdown in the invoke macro where I could select a macro instead of typing it. If there should also be an option of typing, we could introduce a combo box, which allows typing and dropdown selection
* it would be great if all nodes inside the macro would move with it when the macro is being moved
* when right clicking the macro node it says add output. this is a bit confusing as you are creating input parameters
  * can a macro have more than one output?
* with this layout we could think about making the macro node collapsible

magnificus:
Will try to implement improvements.

Regarding input/output, it is that way on purpose, as data is coming as "output" into the stuff inside the macro being inputted when called, do you think this should be explained some other way?

fakob:
You are right. Logically it is an output you are creating. To me it feels a bit confusing though. I was expecting to create "input parameters" which I can input data to when invoking. We could make an exception in the context menu for the macro node and change it to something like "Add input parameter" or "Expose input" or so?

In the old Softimage they had a related feature, which might be even better as we could avoid the whole issue. In a Softimage compound, you always have 1 socket (placeholder), which is not an actual socket.
* You drag from it to connect to another socket
* After connecting, the actual input socket gets created and the socket you dragged from stays so you can create more inputs from it.
See here: https://youtu.be/do1kzR6gVYk?t=1109

This behaviour we could also do for other nodes which allow to add more inputs or outputs like the Add node. It shows 2 regular inputs and one placeholder input. As soon as you connect something, it creates a regular input keeping the placeholder. The added sockets could then also automatically disappear, as you can easily connect them back via the placeholder one.

Screenshot 2022-11-18 at 21 48 06

fakob:
One more thing. The name "Invoke macro" feels very technical. What do you think about renaming it to "Execute macro"?

magnificus:
Looks interesting I will look at that parameter behaviour

PR #182, opened 2022-11-19
Add dropdown widget

fakob:
* Added a dropdown widget with single and multi select option
* Added label option to all widgets

Screenshot 2022-11-19 at 21 20 56

### DIFF
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -57,6 +57,7 @@ export default class PPNode extends PIXI.Container {
   nodePosY: number;
   nodeWidth: number;
   nodeHeight: number;
+  initialData: any;
 
   updateBehaviour: UpdateBehaviourClass;
   nodeSelectionHeader: NodeSelectionHeaderClass;
@@ -207,6 +208,7 @@ export default class PPNode extends PIXI.Container {
     this.y = customArgs?.nodePosY ?? 0;
     this.nodeWidth = this.getDefaultNodeWidth();
     this.nodeHeight = this.getDefaultNodeHeight(); // if not set height is defined by in/out sockets
+    this.initialData = customArgs?.initialData;
     this.isHovering = false;
 
     const inputNameText = new PIXI.Text(
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -57,7 +57,6 @@ export default class PPNode extends PIXI.Container {
   nodePosY: number;
   nodeWidth: number;
   nodeHeight: number;
-  initialData: any;
 
   updateBehaviour: UpdateBehaviourClass;
   nodeSelectionHeader: NodeSelectionHeaderClass;
@@ -208,7 +207,6 @@ export default class PPNode extends PIXI.Container {
     this.y = customArgs?.nodePosY ?? 0;
     this.nodeWidth = this.getDefaultNodeWidth();
     this.nodeHeight = this.getDefaultNodeHeight(); // if not set height is defined by in/out sockets
-    this.initialData = customArgs?.initialData;
     this.isHovering = false;
 
     const inputNameText = new PIXI.Text(
--- a/src/nodes/widgets/widgetNodes.tsx
+++ b/src/nodes/widgets/widgetNodes.tsx
@@ -20,6 +20,7 @@ import {
   SelectChangeEvent,
   Switch,
   ThemeProvider,
+  Typography,
 } from '@mui/material';
 import ColorizeIcon from '@mui/icons-material/Colorize';
 import { SketchPicker } from 'react-color';
@@ -48,7 +49,7 @@ const stepSizeName = 'Step size';
 const maxValueName = 'Max';
 const offValueName = 'Off';
 const onValueName = 'On';
-const buttonTextName = 'Button text';
+const labelName = 'Label';
 const optionsName = 'Options';
 const selectedOptionName = 'Selected option';
 const multiSelectName = 'Select multiple';
@@ -74,13 +75,7 @@ export class WidgetButton extends HybridNode {
     return [
       new Socket(SOCKET_TYPE.IN, offValueName, new AnyType(), 0, false),
       new Socket(SOCKET_TYPE.IN, onValueName, new AnyType(), 1, false),
-      new Socket(
-        SOCKET_TYPE.IN,
-        buttonTextName,
-        new StringType(),
-        'Button',
-        false
-      ),
+      new Socket(SOCKET_TYPE.IN, labelName, new StringType(), 'Button', false),
       new Socket(SOCKET_TYPE.OUT, outName, new AnyType()),
     ];
   }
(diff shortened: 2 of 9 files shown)


src/nodes/widgets/widgetNodes.tsx
```diff
@@ -767,9 +797,275 @@ export class WidgetSlider extends HybridNode {
...
+  public getDefaultNodeHeight(): number {
+    return 104;
+  }
+
+  constructor(name: string, customArgs?: CustomArgs) {
+    super(name, {
+      ...customArgs,
+    });
+
+    if (customArgs?.initialData) {
```
magnificus, line 882: this looks more generic than for this node, should this be in the nodeclass? initialdata, I imagine this is copy pasted in other nodes as well

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -767,9 +797,275 @@ export class WidgetSlider extends HybridNode {
...
+      console.log(customArgs?.initialData);
```
magnificus, line 883: also remove log

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -767,9 +797,275 @@ export class WidgetSlider extends HybridNode {
...
+      this.setInputData(optionsName, customArgs?.initialData);
+    }
+
+    // when the Node is added, add the container and react component
+    this.onNodeAdded = () => {
+      const options = this.getInputData(optionsName) || defaultOptions;
```
magnificus, line 889: remove "||" and instead make sure this.getInputData(optionsName) always returns something valid

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -767,9 +797,275 @@ export class WidgetSlider extends HybridNode {
...
+      ...customArgs,
+    });
+
+    if (customArgs?.initialData) {
+      console.log(customArgs?.initialData);
+      this.setInputData(optionsName, customArgs?.initialData);
+    }
+
+    // when the Node is added, add the container and react component
+    this.onNodeAdded = () => {
```
magnificus, line 888: rewrite this as public onNodeAdded(): void { etc.

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -767,9 +797,275 @@ export class WidgetSlider extends HybridNode {
...
+        },
+        {
+          overflow: 'visible',
+        }
+      );
+      this.setOutputData(outName, options);
+      super.onNodeAdded();
+    };
+
+    this.update = (): void => {
```
magnificus, line 914: rewrite this as public update(): void {

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -767,9 +797,275 @@ export class WidgetSlider extends HybridNode {
...
+    };
+
+    // when the Node is loaded, update the react component
+    this.onConfigure = (): void => {
+      this.update();
+      this.setOutputData(outName, this.getInputData(selectedOptionName));
+      this.executeOptimizedChain();
+    };
+
+    this.onNodeResize = () => {
```
magnificus, line 939: same for these, dont save down a callback function into a variable, instead just define them on the base object like normal functions

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -767,9 +797,275 @@ export class WidgetSlider extends HybridNode {
...
+      this.update();
+    };
+
+    this.onExecute = async function () {
```
magnificus, line 943: same here etc. etc.

magnificus (changes requested):
Very cool, just some things on the structure of the node

src/nodes/widgets/widgetNodes.tsx
fakob: This I had added because I can't control what the user inputs or another node inputs. Otherwise I would have to handle it in the react component. Is that better?

magnificus: Wouldnt it always be an array? is it an arraytype? If so, can add parsing in the arrayType.parse

fakob: I guess that is a general question. You are right, we can fix this for the arraytype, but if the user changes the type to something else, which is possible, then it could potentially be undefined.
Should we not make sure that a node can handle undefined independently?

magnificus: I think it is better to not handle null here, if so then every node has to do it, there is a parse function on the type which can be used to transform even null to for example an empty array, its on the datatypeclass

src/nodes/widgets/widgetNodes.tsx
fakob: I have now moved it into the nodeclass. This required changes also in other nodes. Mainly hybridnodes.

src/nodes/widgets/widgetNodes.tsx
fakob: Done

src/nodes/widgets/widgetNodes.tsx
fakob: If this is what we always want, we should then update the parsing function for all types to always give a valid value. 
Should I do that in this task?

src/nodes/widgets/widgetNodes.tsx
fakob: done

src/nodes/widgets/widgetNodes.tsx
fakob: done

src/nodes/widgets/widgetNodes.tsx
fakob: done

src/nodes/widgets/widgetNodes.tsx
fakob: done

src/classes/NodeClass.ts
```diff
@@ -57,6 +57,7 @@ export default class PPNode extends PIXI.Container {
   nodePosY: number;
   nodeWidth: number;
   nodeHeight: number;
+  initialData: any;
```
magnificus, line 60: we don't have to persist this initialdata do we? Just use in constructor then disregard

magnificus (approved):
It looks good I think, apart from persisting intialdata, can we remove it as a field on node? Functions do not look like normal function definitions but won't harass you about that, thanks for doing this!

src/nodes/widgets/widgetNodes.tsx
magnificus: don't have to do it in this task if you don't want, as long as we plan for it :)

src/classes/NodeClass.ts
fakob: Maybe I misunderstood, but I was basing this change on this comment:
> this looks more generic than for this node, should this be in the nodeclass? initialdata, I imagine this is copy pasted in other nodes as well

I can change it back again, having the initialData only on the node instances which need it. The persisting of data is necessary for hybrid nodes though. Also, this would mean that I have to change some of the public onNodeAdded functions back into the constructor again.

magnificus: I see, I think we are struggling because of some structural issue here, what about adding it to the hybridnode only?

PR #183, opened 2022-11-27
new funcionality and macro polish

magnificus:


fakob (approved):

src/nodes/datatypes/triggerType.tsx
```diff
@@ -39,7 +39,6 @@ export class TriggerType extends AbstractType {
   };
 
   onDataSet(data: any, socket: Socket): void {
```
fakob, line 41: @magnificus could you have a look at this commit again? 
I noticed that the trigger sockets stopped working and I believe it must be because onDataSet is not called anymore.
I attached an example graph you could test with.
[trigger example - 2022-12-02 at 00.22.38.ppgraph.zip](https://github.com/fakob/plug-and-play/files/10136892/trigger.example.-.2022-12-02.at.00.22.38.ppgraph.zip)

PR #184, opened 2022-11-30
InterfaceController, separate model from view

magnificus:


magnificus:
certain regression risk

fakob (approved):

PR #185, opened 2022-11-30
Fix json output for table node

fakob:
I noticed that I had made a mistake on the JSON output of the table node. It was wrongly outputting the raw sheet data. Sorry that I have only noticed it after you had written the extra nodes.

* I have changed the JSON output to actually output JSON. As this is the same as what your Table_GetRowObjects node had done, this node was not needed anymore.
  * The raw sheet data is already included in the workbook output, in case this is ever needed
* I have changed your accompanying nodes to be based on the ArrayOfArrays output which made them a bit simpler


magnificus:
oh nice

magnificus (approved):

PR #188, opened 2022-12-03
Add debounce node

fakob:
The node has all parameters the lodash debounce function has.
If you change the parameters, you need to call the update trigger. This creates a new debounce function with the new values. Having to press the update trigger button is not very elegant, but I was not sure how else to do this.

Screenshot 2022-12-03 at 22 34 16

### DIFF
--- a/src/nodes/utility/utility.ts
+++ b/src/nodes/utility/utility.ts
@@ -184,9 +184,14 @@ export class ThrottleDebounce extends PPNode {
     ];
   }
 
-  // when the Node is added, define debounced function
+  // when the Node is added (new node), define debounced function
   public onNodeAdded = () => {
     super.onNodeAdded();
+    this.updateDebounceFunction();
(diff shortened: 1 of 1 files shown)


magnificus:
Im pretty sure you can do this with a dynamic enum and then not having to use the onAdded function, but will approve nonetheless

magnificus (approved):
comment

fakob:
Not sure how to use the dynamic enum in this case.
The issue I am trying to solve is to only update the debounce function if some of the input parameters change (wait, maxWait, leading, trailing).

However, regardless of this, I still need to initialise the debounce function once
* for a newly added node (onNodeAdded)
* for a loaded node (onConfigure) -> see updated PR

fakob:
> couldnt you just initialize it the first time it is executed?
ah sure. that is a good idea.
Done

PR #189, opened 2022-12-03
add open url node and small fixes

fakob:
* Added an open url node
* Fixed an issue where folders in the examples repo would be listed in the graph search


magnificus (approved):
Looking good

PR #190, opened 2022-12-03
Allow mouse dragging around anything drawable (unless class blocks it specifically)

magnificus:
more polish to be added, but pretty happy with this conceptually

### DIFF
--- a/src/nodes/draw/abstract.tsx
+++ b/src/nodes/draw/abstract.tsx
@@ -23,6 +23,7 @@ export const scaleYName = 'Scale Y';
 export const inputRotationName = 'Angle';
 export const inputPivotName = 'Pivot';
 export const inputAbsolutePositions = 'Absolute Positions';
+export const inputAlwaysDraw = 'Always Draw';
 export const injectedDataName = 'Injected Data';
 export const outputPixiName = 'Graphics';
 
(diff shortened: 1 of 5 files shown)


src/nodes/draw/abstract.tsx
```diff
@@ -127,8 +137,46 @@ export abstract class DRAW_Base extends PPNode {
...
+      this.setOffsets(
+        new PIXI.Point(originalOffsets.x + diffX, originalOffsets.y + diffY)
+      );
+      // we dont re-trigger immediately, to save perfomance
+      setTimeout(
+        () => this.pointerDown(originalCursorPos, originalOffsets),
+        16
+      );
+    });
+    this.deferredGraphics.on('pointerup', () => {
```
fakob, line 164: Maybe it is my touchpad, but sometimes when I release (pointerup) the shape continues moving. Then I have to click and release a second time.

fakob:
Nice work

fakob (approved):
I saw that you can setOffsets by clickin on the drawing and move it. 
Not sure I understood what inputAlwaysDraw does.

src/nodes/draw/abstract.tsx
magnificus: Yes its a bug ATM because it only listens to up events on itself, going to add a new listener to address this

PR #191, opened 2022-12-03
fix trigger socket issues

fakob:
* Fixes issue that socketShouldAutomaticallyAdapt being applied to triggerSockets?
* Small styling change of trigger Socket widget

### DIFF
--- a/src/classes/SocketClass.ts
+++ b/src/classes/SocketClass.ts
@@ -159,7 +159,7 @@ export default class Socket extends PIXI.Container {
     this._data = newData;
     if (
       this.getNode()?.socketShouldAutomaticallyAdapt(this) &&
-      this.dataType.constructor !== new TriggerType().constructor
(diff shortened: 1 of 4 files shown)


src/classes/SocketClass.ts
```diff
@@ -157,7 +157,10 @@ export default class Socket extends PIXI.Container {
   // for inputs: set data is called only on the socket where the change is being made
   set data(newData: any) {
     this._data = newData;
-    if (this.getNode()?.socketShouldAutomaticallyAdapt(this)) {
+    if (
```
magnificus, line 160: would be cleaner to for example ask datatype instead (then triggertype can override the response)

magnificus (approved):

src/classes/SocketClass.ts
fakob: Ah, yes. done

PR #192, opened 2022-12-03
Add row menu

fakob:
* Added right click option to add rows above/below and remove row

Screenshot 2022-12-04 at 00 26 24


magnificus (approved):

fakob:
Also added the ability to right click in the header

PR #194, opened 2022-12-04
made node browser not appear if old link had input

magnificus:
This has been annoying me for some time, when you disconnect a link by dragging output first, it pops up browser, made it only happen if there is no input from before


fakob (approved):
Thank you

PR #195, opened 2022-12-04
socket highlight matching

magnificus:


fakob (approved):
Another one. Very nice!

Do you think it would make sense to give the socket container a 0.01 transparent background. This way we do not have the gap between the socket and the text, which is awkward when right clicking or dragging. Can be done separately though.

https://user-images.githubusercontent.com/4619772/205504549-3a9f7b96-c680-4ad2-af7a-6a2634e265e2.mov

magnificus:
ah right, can look

PR #196, opened 2022-12-04
Playground node

fakob:
* Added a playground node
  * Add all nodes
  * Output graph JSON
  * Output all added nodes
  * Zoom to fit selected nodes
  * Trigger graph method
  * Trigger node method
* Made first updates to the node grouping/coloring
  * The default node color is now the transform color as this is the majority of our nodes
  * The others need to set a color based on the NODE_TYPE_COLOR list

For testing the node consistency including colors I have created a playground node. There are some functions exposed out of the box and I also allow to trigger graph or node functions with no arguments. There are only a few functions which really make sense to call, but I wanted to give it a try. It is not super pretty, as it is mainly for testing.

Screenshot 2022-12-04 at 19 44 48


src/nodes/widgets/widgetNodes.tsx
```diff
@@ -27,10 +27,11 @@ import { SketchPicker } from 'react-color';
...
 import { CustomArgs, TRgba } from '../../utils/interfaces';
 import {
+  NODE_TYPE_COLOR,
   PRESET_COLORS,
+  RANDOMMAINCOLOR,
   SOCKET_TYPE,
   customTheme,
-  RANDOMMAINCOLOR,
 } from '../../utils/constants';
 import { roundNumber } from '../../utils/utils';
```
magnificus, line 36: Maybe there should be a class WidgetNode that extends that then all of these extends? then wouldnt have to specify color for them each manually

magnificus (approved):
Looks very interesting

src/nodes/widgets/widgetNodes.tsx
fakob: Done

PR #197, opened 2022-12-06
Fix issue when removing mouse node

fakob:
This fixes an issue when removing the mouse node. I was overwriting `onNodeDragOrViewportMove` which existed on the nodeclass and that created an issue when the mouse node unsubscribed the one from the nodeclass.


magnificus:
My take on this node is to not use that listener at all but instead on tick call getCursorPosition(), no subscription/unsubscription neccessary

magnificus (approved):

fakob:
Sounds good. Will change.

PR #198, opened 2022-12-09
different rescale, good idea or nah?

magnificus:


fakob:
Interesting idea. I am wondering what problem you are trying to solve? Making some nodes larger based on their importance? Should it work the same way for hybrid nodes? 
I have not seen such a behaviour in node based environments before, which does not mean that it shouldn't be done. Importance, simplification, grouping are are all ways of directing the user's attention. Sometimes done using grouping nodes oder expandable macros.

PR #199, opened 2022-12-14
Load npm node

fakob:
Added a convenience node for npm loading.
The console log inside the custom function is for listing the supported methods in the console. That was the easiest for now.

Screen Shot 2022-12-14 at 22 25 06


magnificus (approved):
looking good

PR #200, opened 2022-12-17
shrink on socket remove

magnificus:


fakob (approved):
Does this also shrink/expand the node when hiding/showing a socket?

magnificus:
now it does

PR #201, opened 2022-12-17
Segments

magnificus:


fakob (approved):
Thanks for implementing segments! I had been wondering how to do it myself. This is a really good feature.
Regarding the json extension, I would probably suggest either the same extension we have for graphs, or a separate one (referring to the discord discussion)?
Regarding adding it to the canvas. Do you think it is possible to add, in this case of a segment being a macro, an execute macro as well right away?
Regarding the simpleGraphBar:
* I tried it out myself, but did not get it to work. I was wondering if you could add example data into the sockets where, when adding an execute macro, it would already show maybe 3 bars?
* Is it possible to name the parameters, so the user knows what to plug in?
* Is it possible that the execute macro shows all parameters when switching macros. I had to execute it once so it actually shows the sockets

magnificus:
Yes there is something wrong with this graph, will need to investigate, default should show bars yes

magnificus:
will polish up

magnificus:
parameter names in macros is a longer discussion, lets talk about it maybe after work

PR #204, opened 2022-12-31
Painfully pulled UI out of graph loading backend and vice versa

magnificus:


magnificus:
There is more work to do but this is a reasonable milestone

fakob (approved):
Thanks for the work! That does indeed look painful.

PR #205, opened 2023-01-03
upgrade to new yarn

fakob:


magnificus (approved):
.cache push to git intentional?

fakob:
I used the gitignore options suggested here when using zero-installs. I think this is the point of zero installs if I understand correctly.
https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

PR #208, opened 2023-01-08
All custom nodes display whether or not they are modified, dedicated look for simple math operations

magnificus:


src/nodes/data/dataFunctions.tsx
```diff
@@ -156,6 +163,7 @@ function getFunctionFromFunction(inputFunction: string): string {
 
 // customfunction does any number of inputs but only one output for simplicity
 export class CustomFunction extends PPNode {
+  modifiedBanner: PIXI.Graphics;
```
fakob, line 166: Is there a way to not show this for a default custom function node, but just for nodes which instantiate the custom function? A default custom function node will always show modified as the default code will always be changed.

fakob (approved):

PR #210, opened 2023-01-08
Force socket to be visible

fakob:


src/classes/GraphClass.ts
```diff
@@ -567,6 +567,13 @@ export default class PPGraph {
     const preTargetName = input.name;
     const preTargetNodeID = input.getNode().id;
 
+    // force socket to be visible
```
magnificus, line 570: I think a socket already is forced to be visible when it has a link, isnt it? If not there might be a bug there and would probably be even better to fix there than to add this adidtional step

fakob:
Ah yes, you are right. I hadn't seen that.
I have now fixed the bug. The setVisible has to happen a bit earlier and
* I added the ifs as otherwise it would force unnecessary rerenders
* I added output as there also is a case for connecting to a hidden output

magnificus (approved):

PR #214, opened 2023-01-15
Improve inspector part 2

fakob:
Based on your comments, this is another update to the inspector with the following features:
* the properties of a node can now be filtered using "tabs"
  * Common
    * also visible if there are multiple nodes selected
    * currently it only contains update behaviour, but could in the future also show color, size, type, or other configs
  * Triggers - trigger sockets - only visible if there is a trigger socket
  * In - input sockets
  * Out - output sockets
  * Source - node config and class code
    * Regarding node config
      * added a "Replace" button , which allows you to edit the config source and replace. I did not know how to keep the id, so the updated node gets a new id
      * added a "New" button which just creates a new node without replacing
* rename the node editing the name in the header

I also made some other small changes
* code editors in the inspector now only load the first 1000 characters. this might reduce the loading weight and reduces also scrolling when scrolling through the sockets in the inspector
* removed the floating node menu as all functionality is now available in the inspector
* added a deselect node short cut - ctrl-shift-a
* made the text on the color widget use a contrast color

https://user-images.githubusercontent.com/4619772/212568141-9adbdb5b-3291-44e5-a0b5-c925ef247108.mov

Screenshot 2023-01-15 at 22 24 21


src/PropertyArrayContainer.tsx
```diff
@@ -1,106 +1,312 @@
...
   sockets: Socket[],
   props: PropertyArrayContainerProps,
-  text: string
-) => {
+  text: string,
+  filter: string,
+  value: string
+) {
   {
     return (
```
magnificus, line 179: bunch of checks here, a bit ugly, why would filter be null? cant it be something else instead? Also, why would sockets be null? it should be [] if there are no sockets

src/PropertyArrayContainer.tsx
```diff
@@ -1,106 +1,312 @@
...
   PropertyArrayContainerProps
 > = (props) => {
   const [dragging, setIsDragging] = useState(
-    PPGraph.currentGraph.selection.isDraggingSelection
+    false
+    // PPGraph.currentGraph.selection.isDraggingSelection
   );
+
+  const singleNode =
+    props.selectedNodes.length > 0 ? props.selectedNodes?.[0] : null;
```
magnificus, line 305: selectednodes should never be null, dont need to do ?.

src/PropertyArrayContainer.tsx
```diff
@@ -111,278 +317,165 @@ export const PropertyArrayContainer: React.FunctionComponent<
...
-                  }}
-                >
-                  {entry}
-                </MenuItem>
-              );
-            })}
-        </Menu>
+          {(props.selectedNodes.length !== 1 ||
+            props.filter === 'common' ||
+            props.filter == null) && (
```
magnificus, line 420: once again would really like to avoid nullfilter, instead do a filter that allows everything through as default

src/PropertyArrayContainer.tsx
```diff
@@ -111,278 +317,165 @@ export const PropertyArrayContainer: React.FunctionComponent<
...
+                'in'
+              )}
+              {socketArrayToComponent(
+                selectedNode.outputSocketArray,
+                props,
+                'Outputs',
+                props.filter,
+                'out'
+              )}
+              {(props.filter === 'source' || props.filter == null) && (
```
magnificus, line 454: same here etc etc

src/classes/GraphClass.ts
```diff
@@ -474,25 +474,55 @@ export default class PPGraph {
...
     const newNode = this.addSerializedNode(
-      oldSerializedNode,
+      newSerializedNode ?? oldSerializedNode,
       {
         overrideId: newId,
       },
       newType
     );
-    if (newType) {
+    if (newType && newSerializedNode === undefined) {
```
magnificus, line 520: why do we need to check newSerializedNode against undefined here? its not used in expression

src/components/ResponsiveDrawer.tsx
```diff
@@ -108,10 +110,12 @@ const ResponsiveDrawer = (props) => {
           randomMainColor={props.randomMainColor}
           handleDrawerToggle={handleDrawerToggle}
         />
-        {props.selectedNode ? (
+        {props.selectedNodes?.length > 0 ? (
```
magnificus, line 113: replace with just props.selectedNodes.length (dont need to have a comparator)

src/InspectorContainer.tsx
```diff
@@ -1,16 +1,146 @@
...
+      }}
+    >
+      <Box
+        sx={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          alignItems: 'center',
+        }}
+      >
+        {props.selectedNodes.length === 1 ? (
```
magnificus, line 34: this expression is very big, can it be replaced with two functions instead maybe?

src/InspectorContainer.tsx
```diff
@@ -1,16 +1,146 @@
...
+          <Typography
+            sx={{
+              pl: 1,
+              py: 0.5,
+            }}
+          >
+            {`${props.selectedNodes.length} nodes selected`}
+          </Typography>
+        )}
+        {props.selectedNodes.length === 1 && (
```
magnificus, line 89: this and the logical expression below have a lot in common, combine them to the extent its possible

magnificus (changes requested):
Looks very promising (especially glad about red lines), but Im hoping we can lose some of the many branching conditionals we have now, the more conditions and if's the higher the risk one of them leads to bugs

src/PropertyArrayContainer.tsx
fakob: The ToggleButtonGroup MUI component returns null in this setup.

Screenshot 2023-01-16 at 21 47 31

fakob: Ah, yes you are right, the socket should always be [].

src/PropertyArrayContainer.tsx
fakob: As mentioned above. Null is what I get back from the component. I can map the null value to another value, for example, "showAll", but I would still need to check for either "common" or "showAll". Is that better?

src/classes/GraphClass.ts
fakob: newSerializedNode is a new optional parameter I have added. And only if you are passing in a newType, but not if you pass in a newSerializedNode, you should rename the node to its type. So in essence you can either replace something using
* a newType
* OR a newSerializedNode
Would it be better to only use one optional parameter? `new?:string | SerializedNode`? and then check if it is a string or a SerializedNode doing logic based on that?

src/components/ResponsiveDrawer.tsx
fakob: Done

src/PropertyArrayContainer.tsx
fakob: done

src/InspectorContainer.tsx
fakob: Good catch. Done.

src/InspectorContainer.tsx
fakob: Splitting them apart as of above request, helped to also make this cleaner.

magnificus (approved):

PR #215, opened 2023-01-22
Added share via Github gist option

fakob:
To share a playground you have to
1. Click on share in the top right corner
2. Click on Login with Github if you are not already logged in
3. Enter an optional description, edit the filename and choose the gist to be secret or public (when selecting secret, the gist will not show up publicly, the link will always work)
4. Click Share playground
5. We create a gist and present
  * an option to copy a shareable link
  * a link to the gist where we also added the shareable link to the description

Regarding oauth
* I am storing the sessions in an in memory store on the server. This means that on redeploy, logged in users will have to log in again
* On the client side, we are storing a session cookie when the user logs in

Known limitations
* Currently when logging in, the page is reloaded. In case you have unsaved changes, they will be lost.

https://user-images.githubusercontent.com/4619772/213940494-fdf24cb8-d750-44d8-a1be-96626872109a.mov


src/App.tsx
```diff
@@ -305,6 +308,20 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
   // on mount
   useEffect(() => {
+    (async function () {
+      const res = await fetch('/api/me', {
+        credentials: 'include',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      });
+      console.log(res);
```
magnificus, line 318: lets remove log

src/App.tsx
```diff
@@ -305,6 +308,20 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
+      const { sessionExpired } = await res.json();
+      if (!sessionExpired) {
+        setIsLoggedIn(true);
+      }
+    })();
+
     console.log(pixiContext.current);
```
magnificus, line 325: also this log

magnificus (approved):

PR #217, opened 2023-01-29
Companion pnp side

magnificus:


src/PropertyArrayContainer.tsx
```diff
@@ -23,6 +23,7 @@ import Socket from './classes/SocketClass';
 import { SocketContainer } from './SocketContainer';
 import { CodeEditor } from './components/Editor';
 import InterfaceController, { ListenEvent } from './InterfaceController';
+import useInterval from 'use-interval';
```
fakob, line 26: I don't see a change in the package.json. Is this package already included?

src/PropertyArrayContainer.tsx
```diff
@@ -397,6 +413,14 @@ export const PropertyArrayContainer: React.FunctionComponent<
     });
   };
 
+  const [socketsCurrentlyRendered, setSocketsCurrentlyRendered] = useState(
+    getSocketsCurrentlyRendered(singleNode)
+  );
+
+  useInterval(() => {
+    setSocketsCurrentlyRendered(getSocketsCurrentlyRendered(singleNode));
+  }, 100);
```
fakob, line 422: Do I understand correctly that every 100ms it checks and updates?

fakob (approved):

src/PropertyArrayContainer.tsx
magnificus: yes indeed

src/PropertyArrayContainer.tsx
magnificus: yes indeed :)

PR #218, opened 2023-01-29
Socket click to focus

fakob:
* When clicking on a socket
  * the inspector opens
  * switches to the respective filter
  * scrolls to and marks the socket
* moved the toast messages to appear in the center, to not cover the drawer button
* moved the share button to the top left

https://user-images.githubusercontent.com/4619772/215353983-baccb990-f80a-4988-9960-c49baa60cf0e.mov

### DIFF
--- a/src/PropertyArrayContainer.tsx
+++ b/src/PropertyArrayContainer.tsx
@@ -169,6 +169,7 @@ function CommonContent(props: CommonContentProps) {
 }
 
 function socketArrayToComponent(
+  socketToInspect: Socket,
   sockets: Socket[],
   props: PropertyArrayContainerProps,
   text: string,
@@ -187,6 +188,7 @@ function socketArrayToComponent(
             {sockets.map((property, index) => {
               return (
                 <SocketContainer
+                  triggerScrollIntoView={socketToInspect === property}
                   key={index}
                   property={property}
                   index={index}
@@ -300,6 +302,9 @@ export const PropertyArrayContainer: React.FunctionComponent<
     false
     // PPGraph.currentGraph.selection.isDraggingSelection
   );
+  const [socketToInspect, setSocketToInspect] = useState<Socket | undefined>(
+    undefined
+  );
 
   const singleNode = props.selectedNodes.length ? props.selectedNodes[0] : null;
   const [selectedNode, setSelectedNode] = useState(singleNode);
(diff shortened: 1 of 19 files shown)


src/PropertyArrayContainer.tsx
```diff
@@ -306,6 +309,20 @@ export const PropertyArrayContainer: React.FunctionComponent<
 
   const [configData, setConfigData] = useState(getConfigData(singleNode));
 
+  function switchFilterBasedOnSelectedSocket(node, socket) {
+    switch (true) {
```
magnificus, line 313: I feel like there is a more elegant way to do this, cant you ask the socket iself what type it is? then it also does not have to be passed into socketArrayToComponent

magnificus (changes requested):
Looks ait and the feature is very useful BUT left a comment on a part in the middle where I think implementation can be different to reduce dependencies and number of lines

src/PropertyArrayContainer.tsx
fakob: I am not sure that actually works. The filter has to be set outside socketArrayToComponent. However, I have now moved the checking in which array the socket is to the socketClass. Please review again.

src/classes/SocketClass.ts
```diff
@@ -301,6 +301,17 @@ export default class Socket extends PIXI.Container {
     return this.parent as PPNode;
   }
 
+  getSocketArrayName(): string {
+    switch (true) {
```
magnificus, line 305: you already know what type the socket is, so no need to go up to the parent to check, check yourself whether you are IN OUT or TRIGGER instead

src/classes/NodeClass.ts
```diff
@@ -619,15 +612,55 @@ export default class PPNode extends PIXI.Container {
...
+    ];
+  }
+  
+  getInputSocketByName(slotName: string): Socket {
+    return this.inputSocketArray[
+      this.inputSocketArray.findIndex((el) => el.name === slotName)
+    ];
+  }
+  
+  getInputOrTriggerSocketByName(slotName: string): Socket {
```
magnificus, line 632: might be problematic if there is a name collision between triggers and input

magnificus (approved):
Looking good!

src/classes/NodeClass.ts
fakob: You are right. Not sure how to solve this though. The Triggers are called Trigger 1, 2 etc. Maybe let's try to not call any normal input Trigger 1?

PR #219, opened 2023-01-30
Fix dropdown/enum options and some cleanup

fakob:
Fixed an issue with the dropdown/enumType where mainly the pivot property in the drawing nodes was not correctly selected after reload. Back in the day I was not consistent in how to handle the dropdown component and what values to use. This I have fixed and have made more consistent.
* Enum/Dropdown options should be an array of objects which have at least a `text` property. No other property is required
* Optionally you can add a `value` property in case you need a separate value. Getting the value would have to be done individually in the node itself
* However when saving, only the `text` property is stored in the socket. An optional `value` would have to be looked up

Also updated the RandomArray node as it had an old school implementation of a node trigger.


magnificus (approved):
nice with more red than green

PR #223, opened 2023-02-03
Unsaved changes warning

fakob:
* Added an unsaved changes warning when
  * closing or reloading a tab/window -> shows standard close/reload tab browser dialog (no control over design or text)
  * loading another graph -> shows the browsers default confirmation dialog (no control over design)
    * for this case we can add a react modal dialog in the future to make it prettier

Screenshot 2023-02-03 at 20 20 29
Screenshot 2023-02-02 at 22 12 20
Screenshot 2023-02-03 at 20 16 05

### DIFF
--- a/src/PPStorage.tsx
+++ b/src/PPStorage.tsx
@@ -14,6 +14,7 @@ import PPGraph from './classes/GraphClass';
 import { hri } from 'human-readable-ids';
 import { Button } from '@mui/material';
 import React from 'react';
+import { SerializedGraph } from './utils/interfaces';
 
 (window as any).__PIXI_INSPECTOR_GLOBAL_HOOK__ &&
   (window as any).__PIXI_INSPECTOR_GLOBAL_HOOK__.register({ PIXI: PIXI });
@@ -22,6 +23,8 @@ import React from 'react';
 const githubBaseURL =
   'https://api.github.com/repos/fakob/plug-and-play-examples';
 const githubBranchName = 'dev';
+const unsavedChangesDialogText =
+  'Changes that you made may not be saved. Do you want to continue?';
 
 // this function is a bit messed up TODO refactor
 function detectTrackPad(event) {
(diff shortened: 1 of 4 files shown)


src/PPStorage.tsx
```diff
@@ -352,35 +379,31 @@ export default class PPStorage {
...
-          >
-            Dismiss
-          </Button>
-        </>
-      ),
-    });
+    if (checkForUnsavedChanges()) {
+      const nameOfFileToClone = remoteGraphsRef.current[id];
+      const fileData = await this.getRemoteGraph(nameOfFileToClone);
+      console.log(fileData);
```
magnificus, line 385: can remove log

src/classes/GraphClass.ts
```diff
@@ -335,6 +336,11 @@ export default class PPGraph {
     }
   }
 
+  onBeforeUnload(event: BeforeUnloadEvent): string {
```
magnificus, line 339: not very clear what this function does, can it be made more explicit?

src/classes/GraphClass.ts
```diff
@@ -48,6 +48,7 @@ export default class PPGraph {
 
   tempConnection: PIXI.Graphics;
   selection: PPSelection;
+  private unsavedChanges: boolean;
```
magnificus, line 51: could this instead be stored inside actionHandler? graphHasUnsavedChanges? cleaner to not expose implicit property this to the graph

src/utils/actionHandler.tsx
```diff
@@ -1,46 +1,50 @@
...
+    action: Action,
+    undo: Action,
+    doPerformAction = true
+  ) {
+    if (doPerformAction) {
       await action();
     }
-    this.undoList.push({action:action, undo:undo});
+    this.undoList.push({ action: action, undo: undo });
+    PPGraph.currentGraph.setUnsavedChange(true);
```
magnificus, line 27: lets keep it inside actionhandler instead if possible

magnificus (changes requested):
Looks good and useful, but would strongly suggest putting the hasChanges variable inside action handler instead, to not let the graph know anything about how it is saved and why etc.

src/classes/GraphClass.ts
fakob: added a comment

src/PPStorage.tsx
fakob: done

src/classes/GraphClass.ts
fakob: moved it to the actionhandler

src/PPStorage.tsx
```diff
@@ -178,41 +200,44 @@ export default class PPStorage {
...
+      try {
+        PPGraph.currentGraph.configure(fileData);
 
-      // unset loadedGraphId
-      await this.db.settings.put({
-        name: 'loadedGraphId',
-        value: undefined,
-      });
+        // unset loadedGraphId
+        await this.db.settings.put({
```
magnificus, line 209: we should have a function "setLoadedGraphID" and "getLoadedGraphID" on PPStorage that does this, so we can hide the name in DB

magnificus (approved):

src/PPStorage.tsx
fakob: done

PR #224, opened 2023-02-05
Some fixes

fakob:
Some small fixes
* fixed array comparison in useInterval
* added useInterval to the codeEditor which now seem to updated properly
* added a character limit to string input
* added drawingOrder to the DRAW_COMBINE_ARRAY
* fixed an issue with the floating json picker
* Screen coordinates of the mouse cursor are now integer

### DIFF
--- a/src/nodes/draw/draw.tsx
+++ b/src/nodes/draw/draw.tsx
@@ -307,10 +307,11 @@ export class DRAW_COMBINE_ARRAY extends DRAW_Base {
       new Socket(SOCKET_TYPE.IN, inputCombineArray, new ArrayType()),
       new Socket(
         SOCKET_TYPE.IN,
-        multiplyYName,
+        numberPerColumnRow,
         new NumberType(true, 0, 100),
         2
       ),
+      new Socket(SOCKET_TYPE.IN, drawingOrder, new BooleanType(), true),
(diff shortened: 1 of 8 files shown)


src/nodes/draw/draw.tsx
```diff
@@ -339,16 +340,25 @@ export class DRAW_COMBINE_ARRAY extends DRAW_Base {
     };
     const myContainer = new PIXI.Container();
     const graphicsArray = inputObject[inputCombineArray];
-    for (let i = graphicsArray.length - 1; i >= 0; i--) {
-      const x = Math.floor(i / inputObject[multiplyYName]);
-      const y = i % inputObject[multiplyYName];
-      const shallowContainer = new PIXI.Container();
-      graphicsArray[i](shallowContainer, executions);
-      shallowContainer.x = x * inputObject[spacingXName];
```
magnificus, line 347: it looks like it would be enough to just flip the coordinates here instead of added a nested loop, or?

magnificus (approved):

src/nodes/draw/draw.tsx
fakob: You are right. Changed this and also the one on the multiplier.

PR #227, opened 2023-02-11
Http node better

magnificus:


src/nodes/api/http.tsx
```diff
@@ -74,40 +73,65 @@ export class HTTPNode extends PPNode {
     ];
   }
 
+  errorColor = TRgba.fromString('#B71C1C');
+  successColor = TRgba.fromString('#4BB543');
```
fakob, line 77: I suggest to move these to constants so we can use the same error and success colors everywhere.

fakob (approved):

src/nodes/api/http.tsx
magnificus: sure

PR #237, opened 2023-03-05
meta input

magnificus:


src/classes/NodeClass.ts
```diff
@@ -297,7 +302,7 @@ export default class PPNode extends PIXI.Container {
     }
 
     // add static inputs and outputs
-    this.getDefaultIO().forEach((IO) => {
+    this.getAllInitialSockets().forEach((IO) => {
       // add in default data if supplied
```
fakob, line 306: Do you want to add meta to all nodes? If so you might want to add it to getDefaultIO as this, instead of getallInitialSockets is used in a lot of places. Or replace the getDefaultIO in the other nodes?

fakob (approved):

src/classes/NodeClass.ts
magnificus: dont want to add it to getDefaultIO directly as it will break anyone not calling the parent and attaching its sockets on top of the parents, this way it will be called for all nodes without having to change anything

PR #242, opened 2023-03-17
Upgrade pixijs to 7.2

fakob:
Now that they had released pixi 7.2 and pixi-viewport, I was able to update them. 
Here a list of the major changes - https://github.com/pixijs/pixijs/releases/tag/v7.2.0

The custom doubleclick script has been removed. Even though they have moved towards supporting the native events, I have implemented doubleclicking using a onClick event and its details property which gives you the count. 

I have tested things, but there might still be some quirks which need to be resolved.


magnificus (approved):
nice

PR #243, opened 2023-03-20
attempt remove loadedgraphid

magnificus:


fakob (commented):
The loadedGraphId was used to load the correct/last graph you were on, when opening the browser. How is this now solved, if it is not stored in the database?

Screenshot 2023-03-20 at 22 54 07

magnificus:
> The loadedGraphId was used to load the correct/last graph you were on, when opening the browser. How is this now solved, if it is not stored in the database?
> 
> Screenshot 2023-03-20 at 22 54 07

was thinking we'd just use the date to open the last modified one, thoughts?

magnificus:
That is the default behaviour and seems to work

fakob:
Ok, sounds good.

PR #244, opened 2023-03-21
fix hybrid load initial load issues

magnificus:


src/classes/NodeClass.ts
```diff
@@ -569,13 +578,6 @@ export default class PPNode extends PIXI.Container {
...
       this.executeOptimizedChain();
     }
-
-    const screenPoint = this.screenPoint();
-    this.onNodeDragOrViewportMove({
-      screenX: screenPoint.x,
-      screenY: screenPoint.y,
-      scale: PPGraph.currentGraph.viewportScaleX,
-    });
   }
```
fakob, line 581: Is this (`refreshNodeDragOrViewportMove`) not needed for a normal node?

fakob (approved):

src/classes/NodeClass.ts
magnificus: hmm yes this should probably stay here

PR #247, opened 2023-04-13
remove seemingly unused dependencies

magnificus:


package.json
```diff
@@ -63,7 +60,6 @@
     "react-frame-component": "^5.2.6",
     "react-lazylog": "^4.5.3",
     "react-monaco-editor": "^0.52.0",
-    "react-responsive-carousel": "^3.2.23",
     "slate": "^0.91.4",
```
fakob, line 63: react-responsive-carousel is a peer dependency of our table package - https://github.com/glideapps/glide-data-grid#-quick-start

fakob (approved):
Other than the commented one, I think it is fine to remove them.

package.json
magnificus: hmm ok

PR #249, opened 2023-04-15
dont execute when graph is closing

magnificus:


fakob (approved):
Nice, thanks for fixing it!

PR #250, opened 2023-04-15
when editing macro, call outside callers

magnificus:


magnificus:
This makes it so that if you are using the macro somewhere, those nodes will be called (if they are changing on update), this makes it a lot easier if you are using them for example to draw graphs

PR #252, opened 2023-04-18
Fix save and rename issue

fakob:
* Fixed the save new and rename issue
* Added option to download any local graph
* Added cut option
* Fixed possible color error for label
* Small search cleanup

### DIFF
--- a/src/PPStorage.tsx
+++ b/src/PPStorage.tsx
@@ -178,16 +178,27 @@ export default class PPStorage {
     }
   };
 
-  downloadGraph() {
-    const serializedGraph = PPGraph.currentGraph.serialize();
-    downloadFile(
-      JSON.stringify(serializedGraph, null, 2),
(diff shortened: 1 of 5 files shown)


src/PPStorage.tsx
```diff
@@ -178,16 +178,27 @@ export default class PPStorage {
...
-      'text/plain'
-    );
-    InterfaceController.showSnackBar(
-      'Current Playground was saved to your Download folder'
-    );
+  async downloadGraph(id = undefined) {
+    this.db
+      .transaction('rw', this.db.graphs, this.db.settings, async () => {
+        const graphId = id || PPGraph.currentGraph?.id;
+        const graph = await this.db.graphs.where('id').equals(graphId).first();
```
magnificus, line 185: this means that if the graph is not saved currently, it will download an old copy of it

magnificus (approved):

PR #253, opened 2023-04-23
Pixijs UI 1

fakob:
* Switched out Button to use PixiUI
* Switched out Slider to use PixiUI
  * I had to change the slider style to something simpler as the PixiUI base was not ideal
  * Removed the Step size socket

https://user-images.githubusercontent.com/4619772/233829464-b401837e-a77f-4c90-95a7-03cf119bc8bb.mov


package.json
magnificus, line 1: Anything that would make sense to remove here now? Import for old nodes stuff?

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -89,81 +115,85 @@ export class WidgetButton extends Widget_Base {
...
-      node.executeChildren();
-    }, []);
-
-    const handleOnPointerDown = () => {
-      node.onWidgetTrigger();
-      const inputData = node.getInputData(onValueName);
-      node.setOutputData(outName, inputData);
-      node.executeChildren();
-    };
+  public onNodeAdded = (source?: TNodeSource) => {
```
magnificus, line 143: I would change how we draw this from onadded to overriding the drawNodeShape function, for examples of this check out macro node or passthrough. This reduces the risk of rendering bugs and edge conditions.

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -89,81 +115,85 @@ export class WidgetButton extends Widget_Base {
...
+
+    super.onNodeAdded(source);
+  };
 
-    const handleOnPointerUp = () => {
-      const inputData = node.getInputData(offValueName);
-      node.setOutputData(outName, inputData);
-      node.executeChildren();
-    };
+  public onNodeResize = (newWidth, newHeight) => {
```
magnificus, line 178: IIRC this should be handled if we put the code in drawNodeShape instead

magnificus: So wouldnt have to copy paste stuff or add custom logic here

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -438,137 +488,149 @@ export class WidgetSlider extends Widget_Base {
     return 104;
   }
 
-  protected getParentComponent(props: any): any {
-    const node = props.node;
-    const [data, setData] = useState(Number(props[initialValueName]));
-    const [minValue, setMinValue] = useState(
-      Math.min(props[minValueName] ?? 0, data)
+  public onNodeAdded = (source?: TNodeSource) => {
```
magnificus, line 491: Same here, lets move stuff out of this function. Do these nodes even need to be hybrid nodes now?

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -438,137 +488,149 @@ export class WidgetSlider extends Widget_Base {
...
+    this.addChild(this._refWidget);
+
+    // Label
+    this._refLabel = new PIXI.Text(
+      String(this.getInputData(labelName)),
+      this.labelTextStyle
     );
-    const [maxValue, setMaxValue] = useState(
-      Math.max(props[maxValueName] ?? 100, data)
+    this._refLabel.anchor.x = 0.5;
```
magnificus, line 537: Similar to above etc.

magnificus (changes requested):
Looks promising but some things i think we should streamline

package.json
fakob: Unfortunately not, we used normal MUI components which both for buttons and sliders are used elsewhere in the overlay UI. However they would be removed via treeshaking in case they are not specifically imported.

magnificus: OK

magnificus (approved):
Looks good if we can would be great to get rid of custom resize logic

PR #255, opened 2023-04-25
new socket dynamic behaviour

magnificus:


fakob (commented):
Very nice. The connecting works really well. Though could you look into unplugging? Before it was possible to take an existing connection and either plug it into another socket, or just let it go. Now it deletes the connection and starts a new one from the input.

PR #256, opened 2023-05-06
Improved search 1

fakob:
* Added groups based on tags/keywords
* Added a _suggested by socket type_ group
  * Only the array type has a preliminary list
* Added a _suggested by node_ group
  * Screenshot 2023-05-04 at 21 46 08
* Added a latest group which shows the 3 last added nodes
  * Screenshot 2023-05-04 at 21 46 28
* Added option to open example graphs for nodes
  * Screenshot 2023-05-04 at 21 47 40
* Merged info with the source tab
  * Shows node description
  * Shows additional description which can contain html, for example links to the companion
  * Screenshot 2023-05-06 at 10 43 31
* Added an edit node icon which opens the inspector on the common tab
  * Screenshot 2023-05-06 at 10 42 22

### Next steps
* Clean up node names
* Update descriptions
  * Add additional descriptions where applicable
* Provide suggested nodes for all data types
* Provide suggested nodes for specific nodes
* Provide node based example graphs for all nodes

https://user-images.githubusercontent.com/4619772/236613646-614da4e6-906f-47cf-81a6-5b360c69881c.mov


src/App.tsx
```diff
@@ -786,80 +780,95 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
...
-          new PIXI.Point(contextMenuPosition[0], contextMenuPosition[1])
-        );
+    if (selected) {
+      const referenceID = hri.random();
+      const addLink = PPGraph.currentGraph.selectedSourceSocket;
+      const setActiveItemArray = () =>
+        setNodeSearchActiveItem((oldArray) => {
+          selected.group = 'Latest';
+          const newArray = [selected, ...oldArray];
+          if (newArray.length > 3) {
```
magnificus, line 790: replace magic number 3 with named constant

src/classes/GraphClass.ts
```diff
@@ -337,18 +337,39 @@ export default class PPGraph {
...
+        {
+          socket: this.socketToInspect,
+          open: shouldOpen,
+        }
       );
     }
   }
 
+  async editNodeMouseDown(node: PPNode): Promise<void> {
+    this.socketToInspect = null;
```
magnificus, line 358: do we need to persist this here on the graph itself? a lot cleaner if we can avoid saving it, especially on PPGraph

src/components/Search.tsx
```diff
@@ -272,15 +272,66 @@ export const getNodes = (): INodeSearch[] => {
...
+      )
+      .sort((a, b) =>
+        a.group?.localeCompare(b.group, 'en', { sensitivity: 'base' })
       );
   }
-  return nodesCached.filter(
-    (node) => !addLink || node.hasInputs
-  ) as INodeSearch[];
+
+  const arrayWithGroupReset = nodesCached.map((node) => ({
```
magnificus, line 288: cache this as well. And also cache a map node names => node, that can be used below in recommendedNodes and preferredNodes to avoid going through the entire list, they just looking the name, goes from time complexity n*m to just m.

src/components/Search.tsx
```diff
@@ -272,15 +272,66 @@ export const getNodes = (): INodeSearch[] => {
...
+    ...node,
+    group: node.tags[0],
+  }));
+
+  const inOrOutputList =
+    (sourceSocket?.isInput()
+      ? sourceSocket?.dataType.recommendedInputNodeWidgets()
+      : sourceSocket?.dataType.recommendedOutputNodeWidgets()) || [];
+
+  const recommendedNodes = inOrOutputList.map((nodeName) => {
```
magnificus, line 298: when you are working on this stuff with the nodes you can just use their names node.key instead of the full object, then at the end you can do a union with no duplicates like this:

var union = [...new Set([...a, ...b])];

Then when you have that list of unique names, you can map them back to their original node names, this should make the code a lot shorter and also faster.

magnificus (changes requested):
Looks super nice, left some comment on the preferred node algorithm that I think can make it simpler

src/App.tsx
fakob: Done

src/components/Search.tsx
fakob: The "latest" group is updated dependent on what the latest nodes are (it mutates the nodes objects). If I would cache arrayWithGroupReset as well, more and more nodes will become latest. That is why I have this reset. Not sure how else to solve that.
Regarding caching the other lists. They are dependent on the sourceSocket so that is always changing. Or did I misunderstand?

src/components/Search.tsx
fakob: I will try this with the node.key. Makes sense.

Regarding using Set: I had used this before, but had seemed to me that it did not keep the order, which is important.

fakob: Actually, regarding node.key, in every map, I do manipulate the group parameter. So I need the object. Or did I misunderstand?

src/classes/GraphClass.ts
fakob: Done. I also moved the socketNameRefMouseDown into the socket class.

src/components/Search.tsx
magnificus: hmm maybe i dont understand what you are doing here exactly, anyway maybe performance is not a problem then it doesnt matter

src/components/Search.tsx
magnificus: alright

magnificus (approved):

PR #257, opened 2023-05-07
Label refactor

magnificus:
It is not perfect but I think its a lot better, there are also some visual bugs fixed


fakob (approved):
Nice!

PR #260, opened 2023-05-07
Updated descriptions and tags

fakob:
* Updated descriptions and tags and some names
* Removed Mandelbrot node


magnificus:
Looks good, some code duplication could be reduced by adding intermediate classes for stuff like InputNode, etc., then they can have the same getTags function and maybe others.

PR #264, opened 2023-05-13
no drag box on phone

magnificus:


fakob (approved):
Nice work!
There are 2 things I am missing, but that does not have to be part of this PR.
* unselect a node when you click somewhere else. otherwise if you add a node the selected node gets replaced
* double click an empty area to get the node search. currently you can click on the logo and click on search nodes, but it would be nice if there is an alternative

magnificus:
yeah there are more things that break on phones we will have to fix

PR #265, opened 2023-05-13
Fix overflow and some mobile adjustments

fakob:
* Fixed issue where the canvas could be extended by overlays adding scroll bars and black background
* when on small width
  * Made inspector go full width
  * Made graph search a bit wide
* for phones, show toasts on the bottom


src/InspectorContainer.tsx
```diff
@@ -45,7 +48,10 @@ function InspectorHeaderReadOnly(props) {
 }
 
 function InspectorHeader(props) {
+  const theme = useTheme();
+  const smallScreen = useMediaQuery(theme.breakpoints.down('sm'));
```
magnificus, line 52: can this be a utility function so it can be reused?

src/components/ResponsiveDrawer.tsx
```diff
@@ -49,6 +51,9 @@ const ResponsiveDrawer = (props) => {
     undefined
   );
 
+  const theme = useTheme();
+  const smallScreen = useMediaQuery(theme.breakpoints.down('sm'));
```
magnificus, line 55: copy pasted

magnificus (approved):
good but lets remove copy pasting

PR #267, opened 2023-05-14
Add arrayToObject node

fakob:


magnificus (approved):
just curious, what are you using this for?

fakob:
Currently with the table node, I get an array of objects. In connection with getting our node descriptions into code I wanted to create an object based json file for the descriptions of our nodes using the node key as key.

PR #272, opened 2023-05-22
Add write to clipboard node

fakob:
* Added a write to clipboard node

Screenshot 2023-05-22 at 23 39 03


magnificus (approved):
interesting

PR #274, opened 2023-05-29
Fix trigger socket disconnect issue

fakob:
* Before, you could click on the socket rect as well as the text to re/disconnect. This interfered with the text clicking which opens the inspector on the right.
* Now, I have limited this to clicking on the socket rect, which also solves this - https://trello.com/c/57zuqE4F

In very rare cases, I have still observed that just clicking on a socket rect, could reconnect it. But I was not able to reproduce it


magnificus:
I think its pretty nice to be able to drag to and from the text if you are butter fingered like me, what about having it behave only this way for trigger inputs?

magnificus (approved):

PR #275, opened 2023-06-01
Add video node

fakob:
* load any video and save it as a local resource
  * videos with same name and size are overwritten
* control the video
  * play/pause
  * loop
  * speed
  * mute
  * volume
  * position in seconds
  * position in percent
* trigger
  * transcoding > uses ffmpeg
  * capture current frame
  * capture an array of frames
    * using an interval
    * using a total amount
* added an example which is accessible through the link in the search node dropdown - see image below

NOTE: currently I have not built in any cleanup for the local resources. I did not want to remove it with the node as there might be some other node with the same resourceId. Have to think about a logic

Screenshot 2023-06-02 at 00 01 00


src/App.tsx
```diff
@@ -262,6 +265,45 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
               }
             }
             break;
+          case '3gp':
```
magnificus, line 268: can we turn this into a separate function and put into another file?

magnificus: to avoid bloating up App.tsx more

magnificus (approved):
Looks cool and good that it is well encapsulated, only point was about App.tsx

PR #276, opened 2023-06-02
Fix add connected widget nodes

fakob:
When right clicking on an input socket text you are able to choose a "widget" node. For the codeEditor and the Label we were carrying over the text from the input socket. Now I have implemented that
* this also works for other widgets like the slider, the trigger, the color picker, ...
* not just data get's carried over but also other properties
  * the socket name in case the widget has a label
  * the min, max, round values in case of the slider widget

Currently this behaviour is only triggered if right clicking the socket name and choosing the widget from their. I wonder if we should make it work in general when dragging a connection from an input socket and choosing an appropriate widget from the search.

Screenshot 2023-06-02 at 21 44 21

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -694,7 +694,7 @@ export default class PPGraph {
         {
           nodePosX: node.x + (socket.isInput() ? 0 : node.width + 40),
           nodePosY: node.y + socket.y,
-          initialData: socket.isInput() ? socket : undefined,
+          initialData: socket.isInput() ? socket.data : undefined,
         },
         NODE_SOURCE.NEWCONNECTED
       );
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -694,7 +694,7 @@ export default class PPGraph {
         {
           nodePosX: node.x + (socket.isInput() ? 0 : node.width + 40),
           nodePosY: node.y + socket.y,
-          initialData: socket.isInput() ? socket.data : undefined,
+          initialData: socket.isInput() ? socket : undefined,
         },
         NODE_SOURCE.NEWCONNECTED
       );
--- a/src/nodes/codeEditor.tsx
+++ b/src/nodes/codeEditor.tsx
@@ -6,8 +6,12 @@ import ErrorFallback from '../components/ErrorFallback';
 import MonacoEditor from 'react-monaco-editor';
 import PPSocket from '../classes/SocketClass';
 import { CodeType } from './datatypes/codeType';
-import { TNodeSource } from '../utils/interfaces';
-import { convertToString, downloadFile, formatDate } from '../utils/utils';
+import {
+  convertToString,
+  downloadFile,
+  formatDate,
+  updateDataIfDefault,
+} from '../utils/utils';
 import { SOCKET_TYPE, customTheme } from '../utils/constants';
 import HybridNode2 from '../classes/HybridNode2';
 
@@ -77,15 +81,9 @@ export class CodeEditor extends HybridNode2 {
   }
 
   public outputPlugged(): void {
-    const links = this.getSocketByName(outputSocketName).links;
-    const target = links[0].getTarget();
-    if (
-      links.length === 1 &&
-      codeDefaultData === this.getInputData(inputSocketName)
-    ) {
-      this.setInputData(inputSocketName, target.defaultData);
-      this.executeOptimizedChain();
-    }
+    const dataToUpdate =
+      this.getSocketByName(outputSocketName).links[0].getTarget().defaultData;
+    updateDataIfDefault(this, inputSocketName, codeDefaultData, dataToUpdate);
     super.outputPlugged();
   }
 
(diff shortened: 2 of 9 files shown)


src/classes/GraphClass.ts
```diff
@@ -690,7 +690,7 @@ export default class PPGraph {
         {
           nodePosX: node.x + (socket.isInput() ? 0 : node.width + 40),
           nodePosY: node.y + socket.y,
-          initialData: socket.isInput() ? socket.data : undefined,
+          initialData: socket.isInput() ? socket : undefined,
```
magnificus, line 693: a socket as initial data sounds crazy, do we need this?

src/nodes/codeEditor.tsx
```diff
@@ -30,6 +31,20 @@ export class CodeEditor extends HybridNode2 {
     return ['Widget'].concat(super.getTags());
   }
 
+  public onNodeAdded = (source?: TNodeSource) => {
```
magnificus, line 34: this looks pretty messy, should we talk about how things are working here?

src/nodes/data/dataFunctions.tsx
```diff
@@ -136,7 +136,14 @@ export class Constant extends PPNode {
 
   public onNodeAdded = () => {
     if (this.initialData) {
-      this.setInputData(constantInName, this.initialData);
+      let data;
```
magnificus, line 139: looks messy and also copy pasted from above

src/nodes/text.tsx
```diff
@@ -97,8 +97,18 @@ export class Label extends PPNode {
 
   public onNodeAdded = (source?: TNodeSource) => {
     if (this.initialData) {
-      this.setInputData('Input', this.initialData);
+      let data;
```
magnificus, line 100: same

magnificus (changes requested):
Hate to say it but this code looks very unhealthy, can we discuss it?

fakob:
I think the feature is very important, but I was also not happy with the solution. I hope we find a better implementation. Let's discuss

magnificus:
Yes, sounds good

Den sön 4 juni 2023 19:20Jakob Schindegger ***@***.***> skrev:

src/nodes/data/dataFunctions.tsx
```diff
@@ -134,12 +141,18 @@ export class Constant extends PPNode {
     return true;
   }
 
-  public onNodeAdded = () => {
-    if (this.initialData) {
-      this.setInputData(constantInName, this.initialData);
+  public outputPlugged(): void {
```
magnificus, line 144: this looks copy pasted from the one above, would make them use the same code

src/nodes/codeEditor.tsx
```diff
@@ -73,6 +76,19 @@ export class CodeEditor extends HybridNode2 {
     return 300;
   }
 
+  public outputPlugged(): void {
+    const links = this.getSocketByName(outputSocketName).links;
+    const target = links[0].getTarget();
+    if (
+      links.length === 1 &&
```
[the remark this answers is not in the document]
fakob: Ah, you are right. Will do

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -355,11 +380,28 @@ export class WidgetRadio extends WidgetBase {
...
+
   public executeOnPlace(): boolean {
     return true;
   }
+
+  public outputPlugged(): void {
+    const target =
+      this.getSocketByName(selectedOptionName).links[0].getTarget();
+    if (
+      target.dataType.constructor === new ArrayType().constructor &&
```
magnificus, line 392: would avoid checking for the class here, in general checking for class is considered an anti-pattern, I would recommend the Array.isArray function

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -397,6 +439,20 @@ export class WidgetColorPicker extends WidgetHybridBase {
     return 104;
   }
 
+  public outputPlugged(): void {
+    const target = this.getSocketByName(outName).links[0].getTarget();
+    if (
+      target.dataType.constructor === new ColorType().constructor &&
```
magnificus, line 445: would recommend something like isColor instead of checking class, consider if it is an AnyType that contains a color for example.

src/nodes/widgets/widgetNodes.tsx
```diff
@@ -863,6 +972,21 @@ export class WidgetDropdown extends WidgetHybridBase {
     return 104;
   }
 
+  public outputPlugged(): void {
+    const target = this.getSocketByName(outName).links[0].getTarget();
+    if (
+      target.dataType.constructor === new ArrayType().constructor &&
```
magnificus, line 978: same comment as above

magnificus (approved):
I like the feature a lot, but added a few more comments

src/nodes/widgets/widgetNodes.tsx
fakob: makes sense

src/nodes/widgets/widgetNodes.tsx
fakob: good idea

PR #277, opened 2023-06-03
Undo interface values

magnificus:


fakob (approved):
The slider works nicely. I also tested label which had some issues if it includes removing it and bringing it back.

1. Added a label
2. made text changes multiple times
3. undid changes including adding the label
4. redoing would bring back the label, but the value changes I could not bring back
The same seems to happen with for example the dropdown.
I assume for color picker and switch this functionality has to be built in separately?

Regarding the input comment (isEventComingFromWithinTextInput). When editing a label, clicking ctrl A for select all, selects all nodes and undo redo does undo redo of the graph, but not inside the label.

PR #279, opened 2023-06-05
Include data of connected sockets

fakob:
* added the ability of keeping the data when copy pasting nodes which have connections which are not copied
* added the execution of all seed nodes when pasting

https://github.com/fakob/plug-and-play/assets/4619772/66abf414-edda-4329-ac20-cafa1e088e17

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -809,6 +809,9 @@ export default class PPGraph {
     this.selection.selectNodes(newNodes, false, true);
     this.selection.drawRectanglesFromSelection();
 
+    // execute all seed nodes to make sure there are values everywhere
+    await this.executeAllSeedNodes(newNodes);
+
     return newNodes;
   }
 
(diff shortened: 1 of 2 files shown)


src/classes/GraphClass.ts
```diff
@@ -988,6 +994,18 @@ export default class PPGraph {
     // get serialized nodes
     const nodesSerialized = nodes.map((node) => node.serialize());
 
+    // include deep copy of data from sockets whos links are not included
+    socketsToIncludeDataOf.forEach((socket) => {
```
magnificus, line 998: i would remove socketsToIncludeDataOf and make this statement instead:

linksContainedInSelection.forEach( link => 
if (link.getTarget().getNode() in selection and link.getSource().getNode() not in selection){
link.getTarget().data = deepcopy
})

It will reduce the code and make it easier to follow

magnificus (approved):
looks good but I suggest changing the logic inside serialization to make it a bit simpler

src/classes/GraphClass.ts
fakob: There must be a misunderstanding. `linksContainedInSelection` are already only the links where all nodes are included. I made a small change to improve understandability.

PR #282, opened 2023-06-11
Load sqlite

fakob:
Added an sqlite reader node which dynamically imports sqlite-wasm
* either drag in a database file (including pxshow)
  * pxshow gets its own SQL query targeting state
* OR specify a URL pointing to a db file
* Add an SQL query
* The node outputs
  * Array of tables
  * SQL column names
  * SQL results


src/dragAndDrop.ts
```diff
@@ -146,6 +148,42 @@ export const dragAndDrop = (acceptedFiles, fileRejections, event) => {
...
+            PPGraph.currentGraph.selection.selectedNodes?.[index]?.type ===
+            'SqliteReader'
+          ) {
+            const existingNode = PPGraph.currentGraph.selection.selectedNodes[
+              index
+            ] as any;
+            existingNode.updateAndExecute(localResourceId, file.path);
+          } else {
+            const sqlQuery =
+              extension === 'pxshow' ? PXSHOW_SQL_QUERY : undefined;
```
magnificus, line 175: maybe cleaner to just always add a new node when dragging in? Makes logic a lot simpler too

src/nodes/utility/database.ts
```diff
@@ -0,0 +1,236 @@
...
+      new PPSocket(SOCKET_TYPE.OUT, outputQuerySocketName, new ArrayType()),
+    ];
+  }
+
+  public onNodeAdded = async (source?: TNodeSource): Promise<void> => {
+    const namePrefix = this.id + ' sqlite-wasm';
+    console.time(`${namePrefix} imported`);
+    const packageName = '@sqlite.org/sqlite-wasm';
+    const url = 'https://esm.sh/' + packageName;
+    this.sqlite3Module = await import(/* webpackIgnore: true */ url);
```
magnificus, line 99: good idea, maybe we should do it this way or similar way for other dependencies?

magnificus (approved):

src/dragAndDrop.ts
fakob: Having the ability to replace a node can be very convenient. I am using this already for images, videos and now also for this. If I always create a new one, then I would have to rewire them. That seems more complicated.

src/nodes/utility/database.ts
fakob: Not all packages are supported for lazy loading, but yes that is something I will try. The next candidate is the XLSX package in the table node. Let's see if it works.

PR #283, opened 2023-06-11
add width to label

fakob:
* Added width to the label
  * 0 or undefined is no width (as wide as text)
  * you can change the width input or resize the node
  * double clicking the resize icon, resets the width to 0
* Added `activeRescaling` parameter in `onNodeResize` to differentiate between actively rescaling with mouse or just calling it from somewhere else
* `onNodeResize` is now only called if the size has changed
* fixed a bug where a new HTMLTextComponent was appended every time one edited the text
  * now all Labels share one single Label-Input div for editing text

https://github.com/fakob/plug-and-play/assets/4619772/4b022a74-447d-4e4f-8064-326a263f0e40

### DIFF
--- a/src/classes/NodeClass.ts
+++ b/src/classes/NodeClass.ts
@@ -79,16 +79,20 @@ export default class PPNode extends PIXI.Container {
   protected statuses: NodeStatus[] = []; // you can add statuses into this and they will be rendered on the node
 
   // supported callbacks
-  onNodeDoubleClick: (event: PIXI.FederatedPointerEvent) => void = () => { };
+  onNodeDoubleClick: (event: PIXI.FederatedPointerEvent) => void = () => {};
   onViewportMoveHandler: (event?: PIXI.FederatedPointerEvent) => void =
-    () => { };
+    () => {};
   onViewportPointerUpHandler: (event?: PIXI.FederatedPointerEvent) => void =
-    () => { };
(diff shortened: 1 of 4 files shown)


src/classes/NodeClass.ts
```diff
@@ -476,15 +480,16 @@ export default class PPNode extends PIXI.Container {
   resizeAndDraw(
     width: number = this.nodeWidth,
     height: number = this.nodeHeight,
-    maintainAspectRatio = false
+    maintainAspectRatio = false,
+    activeRescaling = false // mouse rescaleing
```
magnificus, line 484: this activeRescaling looks like a hack, do we really need it?

magnificus (changes requested):
do we really need to add extra logic to nodeclass to accomodate this? we can resize other nodes and save it already, cant we just have an additional optional socket on label?

magnificus (approved):

src/classes/NodeClass.ts
```diff
@@ -507,7 +515,9 @@ export default class PPNode extends PIXI.Container {
 
     this.nodeSelectionHeader.x = NODE_MARGIN + this.nodeWidth - 96;
 
-    this.onNodeResize(this.nodeWidth, this.nodeHeight);
+    if (oldWidth !== newNodeWidth || oldHeight !== newNodeHeight) {
```
magnificus, line 518: maybe not needed now?

PR #284, opened 2023-06-12
Lazy load XLSX

fakob:
Not a real PR, just to show you how I suggest to do this. Still has wait for being loaded issues when reloading the graph.


magnificus:
I will adopt this branch

fakob (commented):
Apparently I can not approve my own pull request so I will state this:
I am Jakob and I approve this ~message~ PR

PR #285, opened 2023-07-04
Fix shadow issue

fakob:
This should fix the issue you had seen on the code editor. 
When double clicked (in edit mode) I was rendering a shadow in css. This layer was covering the below sockets as well as the resize icon. I have now removed the css shadow and render the shadow on pixi side using a blurred texture.


magnificus (approved):
thanks for this

PR #287, opened 2023-07-08
Macro usability

magnificus:


fakob:
I don't fully understand the code, but it looks good to me. Will this break previous macro examples?

fakob (approved):

magnificus:
This should not break older graphs, but if it does we can revert this and I can look at why

magnificus:
I checked the templates locally and it looked OK

PR #288, opened 2023-07-09
Alt drag to duplicate

fakob:
* Added option to alt+drag to duplicate
* Made pasteNodes undoable

https://github.com/fakob/plug-and-play/assets/4619772/6821de3d-b29f-461e-b53a-f32a968cdc23

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -12,7 +12,11 @@ import {
   TNodeSource,
   TPastePos,
 } from '../utils/interfaces';
-import { connectNodeToSocket, isPhone } from '../utils/utils';
+import {
+  connectNodeToSocket,
+  createArrayOfRandomIds,
+  isPhone,
+} from '../utils/utils';
 import { getNodesBounds } from '../pixi/utils-pixi';
 import PPNode from './NodeClass';
 import PPSocket from './SocketClass';
@@ -519,7 +523,7 @@ export default class PPGraph {
   ) {
     const referenceID = hri.random();
     const action = async () => {
-      PPGraph.currentGraph.replaceNode(
+      await PPGraph.currentGraph.replaceNode(
         oldSerializedNode,
         oldSerializedNode.id,
         referenceID,
@@ -529,7 +533,7 @@ export default class PPGraph {
       );
     };
     const undoAction = async () => {
-      PPGraph.currentGraph.replaceNode(
+      await PPGraph.currentGraph.replaceNode(
         newSerializedNode,
         referenceID,
         oldSerializedNode.id,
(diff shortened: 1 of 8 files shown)


src/classes/GraphClass.ts
```diff
@@ -761,68 +766,82 @@ export default class PPGraph {
     this.allowExecution = true;
   }
 
-  async duplicateSelection(): Promise<PPNode[]> {
+  async duplicateSelection(pastePos: TPastePos = undefined): Promise<PPNode[]> {
```
magnificus, line 769: can we avoid the undefined here? Maybe by making pastepos a mandatory parameter or add existing default value?

src/classes/NodeClass.ts
```diff
@@ -987,14 +987,28 @@ ${Math.round(this._bounds.minX)}, ${Math.round(
...
+      const altKey = event.altKey;
 
       // select node if the shiftKey is pressed
       // or the node is not yet selected
       if (shiftKey || !this.selected) {
         PPGraph.currentGraph.selection.selectNodes([this], shiftKey, true);
         this.onSpecificallySelected();
       }
-      PPGraph.currentGraph.selection.startDragAction(event);
+      if (altKey) {
```
magnificus, line 998: getting a lot of ifs in here becomes a bit hard to digest

src/classes/GraphClass.ts
```diff
@@ -761,68 +766,82 @@ export default class PPGraph {
...
     data: SerializedSelection,
-    pasteTo?: {
-      x: number;
-      y: number;
-    }
+    pastePos?: TPastePos
   ): Promise<PPNode[]> {
     const newNodes: PPNode[] = [];
     const mappingOfOldAndNewNodes: { [key: string]: PPNode } = {};
+    const arrayOfRandomIds = createArrayOfRandomIds(data.nodes.length);
```
magnificus, line 784: cant you just create the random ids when you create the node further down? Can make it less stateful that way, you already have the list of all nodes so you can get their IDS from there for the undo action.

src/utils/utils.ts
```diff
@@ -840,3 +841,11 @@ export const updateDataIfDefault = (
     node.executeOptimizedChain();
   }
 };
+
+export const createArrayOfRandomIds = (length: number): string[] => {
```
magnificus, line 845: dont think we should need this function it seems redundant

magnificus: Also if we need it we can use

Array.from({ length },() => hri.random())

magnificus (approved):
approved but added some comments

src/classes/GraphClass.ts
fakob: This is a similar issue as commented in the other PR. 
* When defined we are doing a general offset based on the pastePos
* When undefined we are doing an individual positioning based on the original nodes position. We are solving the issue downstream when looping through the individual nodes.

src/classes/NodeClass.ts
fakob: I agree, but I am not sure how to solve it. If I move to use a switch I would have to duplicate code as they are not mutually exclusive. What could be another option?

src/classes/GraphClass.ts
fakob: In the undo action I have to delete all newly created nodes. They are deleted using/identified by the random IDs I create here. The IDs from the original nodes would not help me in the undo action as I am not touching them.
Looking at other undoable actions we had always created the Ids for the newly created nodes before. Is there another way to keep track of them?

src/utils/utils.ts
fakob: Thanks for the short version. I was not aware of the Array.from usage.

src/classes/GraphClass.ts
magnificus: what about another function without parameters that calls the second one?

fakob: I am not sure I fully understand. Like this?
```
  duplicateSelectionWithoutPosition() {
    this.duplicateSelection();
  }

  async duplicateSelection(pastePos: TPastePos = undefined): Promise<PPNode[]> {
    const serializeSelection = this.serializeSelection();
    const pastedNodes = await this.action_pasteNodes(
      serializeSelection,
      pastePos
    );
    return pastedNodes;
  }
```
I would still have to pass in undefined? I will merge and we can discuss this further when you are back.

PR #289, opened 2023-07-15
Align nodes

fakob:
* Added options to align and distribute multiple nodes using
  * the icons on the bottom of the selection
  * OR the context menu
* setPosition now also accepts x or y being undefined in case you only want to move on one axis
* cleaned up icon naming
* made Extract to Macro option only show up if there are more than 1 nodes selected

https://github.com/fakob/plug-and-play/assets/4619772/9aa65395-be2b-47da-b895-b1b81f0e4c1d
Screenshot 2023-07-15 at 18 35 47


src/classes/NodeClass.ts
```diff
@@ -461,8 +461,13 @@ export default class PPNode extends PIXI.Container {
   }
 
   public setPosition(x: number, y: number, isRelative = false): void {
-    this.x = isRelative ? this.x + x : x;
-    this.y = isRelative ? this.y + y : y;
+    if (isRelative) {
+      this.x = this.x + (x ?? 0);
```
magnificus, line 465: Instead of this, cant we have "0" as the default parameter value in to the function?

magnificus (approved):

src/classes/NodeClass.ts
fakob: 0 would work well if we always pass relative coordinates, but with absolute ones you need to differentiate:
absolute: setPosition(123, undefined) === setPosition(123, this.y)
relative: setPosition(123, undefined) === setPosition(123, 0)

PR #290, opened 2023-07-16
Add more connect options to socket right click

fakob:
* Right clicking a socket now presents all preferred nodes (not just the first one)
  * custom per node
  * custom per socket type
* This step is now also undoable
* Added Label as a preferred node to the number type

Screenshot 2023-07-16 at 17 15 29

### DIFF
--- a/src/components/ContextMenus.tsx
+++ b/src/components/ContextMenus.tsx
@@ -509,15 +509,8 @@ export const NodeContextMenu = (props) => {
   );
 };
 
-function constructRecommendedNodeOptions(
-  preferredNodesPerSocket: string[],
-  selectedSocket: PPSocket
-): any {
-  const preferredNodes = preferredNodesPerSocket.concat(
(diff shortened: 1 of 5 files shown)


src/components/ContextMenus.tsx
```diff
@@ -507,6 +509,47 @@ export const NodeContextMenu = (props) => {
   );
 };
 
+function constructRecommendedNodeOptions(
+  preferredNodesPerSocket: string[],
+  selectedSocket: PPSocket
+): any {
+  const preferredNodes = preferredNodesPerSocket.concat(
```
magnificus, line 516: maybe this should be its own function somewhere?

magnificus (approved):

src/components/ContextMenus.tsx
fakob: Good idea. Done

PR #293, opened 2023-07-18
Auto font size for button, slider, color picker and switch

fakob:
* Auto font size for button, slider, color picker and switch

Screenshot 2023-07-18 at 23 19 33

compared to

Screenshot 2023-07-18 at 23 20 40


magnificus (approved):
very nice

PR #296, opened 2023-08-02
UI + database refactoring to make things simpler, faster and more future proof

magnificus:


src/PPStorage.tsx
```diff
@@ -281,119 +275,94 @@ export default class PPStorage {
...
-          });
-
-          PPGraph.currentGraph.id = newId;
-
-          InterfaceController.notifyListeners(ListenEvent.GraphChanged, {
-            newId,
-            name,
-          });
-
-          InterfaceController.showSnackBar('New playground was saved');
```
fakob, line 380: Do you think we can add this again? I found it practical.

src/PPStorage.tsx
```diff
@@ -281,119 +275,94 @@ export default class PPStorage {
...
-    updateGraphSearchItems: any
-  ) {
+  renameGraph(graphId: string, newName = undefined) {
     this.db
       .transaction('rw', this.db.graphs, this.db.settings, async () => {
         await this.db.graphs.where('id').equals(graphId).modify({
           name: newName,
         });
-        setActionObject({ id: graphId, name: newName });
-        updateGraphSearchItems();
```
fakob, line 339: Could you try out renaming. When I tried it I got 2 errors:
* from Dexie
* from the MUI graph dropdown
Screenshot 2023-08-03 at 20 37 20

fakob (approved):
Thanks. Looks like a good cleanup. 
Noted a small bug with rename.

src/PPStorage.tsx
magnificus: I will rewrite it as well

src/PPStorage.tsx
magnificus: It should now always trigger, tell me if it doesnt

src/PPStorage.tsx
magnificus: looks ait for me now, tell me if it doesnt work

PR #297, opened 2023-08-05
Add tooltips

fakob:
* Added tooltips when hovering over socket, node, selection
  * Selection node and input socket tooltips are shown on the left side, output sockets are shown on the right side
* When hovering over socket, changed option from ctrl+click to shift+click to pin the inspector (currently only one floating inspector is allowed)
  * Shift click now also works when over the actual socket, not just the text
* Added option to show the node inspector to the socket context menu

https://github.com/fakob/plug-and-play/assets/4619772/3881bcc7-773d-4e5e-94fc-1f53f92a712e

Screenshot 2023-08-05 at 18 55 20
Screenshot 2023-08-05 at 18 56 23
Screenshot 2023-08-05 at 18 56 33


src/classes/NodeClass.ts
```diff
@@ -1086,6 +1087,10 @@ ${Math.round(this._bounds.minX)}, ${Math.round(
   }
 
   onPointerClick(event: PIXI.FederatedPointerEvent): void {
+    PPGraph.currentGraph.viewport.addEventListener(
```
magnificus, line 1090: Can this be done via interfacecontroller instead?

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,168 @@
...
+    isSocket(object) ||
+    (isNode(object) && !(object as PPNode).doubleClicked) ||
+    isSelection(object)
+  );
+}
+
+function Content(props) {
+  const object: TPPType = props.object;
+  switch (true) {
+    case isSocket(object):
```
magnificus, line 33: instead of switch casing in here, could we make an interface for "TooltipBehaviour" that both node and socket implements? (Makes it more scalable and separation of concerns)

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+function shouldShow(object) {
+  return (
+    isSocket(object) ||
+    (isNode(object) && !(object as PPNode).doubleClicked) ||
+    isSelection(object)
+  );
+}
+
+function Content(props) {
```
magnificus, line 30: Can we make this return type static and optional if we need to return null?

function Content(props) : ReactSomething | | undefined

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,168 @@
...
+            {(object as PPSelection).selectedNodes.length} nodes selected
+          </Box>
+          <CodeEditor
+            value={getSelectionTooltipData(selection)}
+            randomMainColor={props.randomMainColor}
+          />
+        </>
+      );
+    default:
+      return null;
```
magnificus, line 106: if possible to avoid null and instead do empty component or something it is more elegant, but no big deal

src/utils/utils.ts
```diff
@@ -840,3 +843,94 @@ export const updateDataIfDefault = (
...
+
+export function isSocket(object) {
+  return (
+    object instanceof PPSocket ||
+    object?.parent instanceof PPSocket ||
+    (object?.parent instanceof PPSocket && object instanceof PIXI.Text)
+  );
+}
+
+export function getSocket(socketish): PPSocket {
```
magnificus, line 875: lets get rid of these and instead work with classes and children, this is very hacky

src/utils/utils.ts
```diff
@@ -840,3 +843,94 @@ export const updateDataIfDefault = (
...
+export function isSelection(object) {
+  return object instanceof PPSelection;
+}
+
+export const getTooltipPositionBasedOnType = (object: TPPType): PIXI.Point => {
+  const scale = PPGraph.currentGraph.viewportScaleX;
+  const distanceX = TOOLTIP_DISTANCE * scale;
+  let absPos;
+  switch (true) {
+    case isSocket(object):
```
magnificus, line 892: yeah same with these

magnificus (changes requested):
Looks very promising but we need to drop the case switch behaviour and work with an interface instead

src/classes/NodeClass.ts
fakob: Done, though I had introduce a new one as I only need `pointerup` and not also `pointerupoutside`

src/components/Tooltip.tsx
fakob: Done

src/components/Tooltip.tsx
fakob: Done

src/utils/utils.ts
fakob: Done

src/utils/utils.ts
fakob: Done

magnificus (changes requested):
Nice, you have understood the idea, you have just implemented it in a very unorthodox way. The normal way to implement it would be to create a new abstract class https://www.tutorialsteacher.com/typescript/abstract-class called something like "TooltipAble", which has a single abstract function "getTooltipContent".

Then on NodeClass and SocketClass you would extend this class, and as the function in "TooltipAble" is abstract, it won't compile until you make a function in them that matches. Then in the check you are doing now, instead of checking for if the function exists, you would check if the object implements "TooltipAble"

magnificus (approved):

PR #298, opened 2023-08-06
Update cypress tests to use node 16

fakob:
For my PR I got a cypress test error. It seems I have to run node 16.

Screenshot 2023-08-06 at 14 47 38


fakob:
There also seems to be an option to do this in runs. Not sure what is better, or how this is different. Do you?
```
runs:
  using: 'node16'
```

magnificus (commented):
LGTM

PR #305, opened 2023-08-14
Fix error stringifying DeferredPixiType

fakob:
This error happened when copy pasting a node which had a DeferredPixiType as input.


magnificus (approved):
nice thanks

PR #310, opened 2023-09-18
Structured node drawing into background and foreground groups

fakob:
I have structured the internals of the node layers a bit and separated them into
* background
  * pretty much the whole node including background, header, sockets
* foreground
  * what ever is rendered, either on top of it (in case of widgets), or outside of it (in case of render nodes)

Screenshot 2023-09-18 at 19 26 49
Screenshot 2023-09-18 at 19 27 05

### DIFF
--- a/src/classes/SocketClass.tsx
+++ b/src/classes/SocketClass.tsx
@@ -337,7 +337,7 @@ export default class Socket extends PIXI.Container implements Tooltipable {
   }
 
   getNode(): PPNode {
-    return this.parent as PPNode;
+    return this.parent?.parent as PPNode;
   }
 
   getGraph(): PPGraph {
--- a/src/classes/UpdateBehaviourClass.ts
+++ b/src/classes/UpdateBehaviourClass.ts
@@ -168,7 +168,7 @@ export default class UpdateBehaviourClass extends PIXI.Container {
   // METHODS
 
   getNode(): PPNode {
-    return this.parent as PPNode;
+    return this.parent?.parent as PPNode;
   }
 
   getGraph(): PPGraph {
--- a/src/classes/UpdateBehaviourClass.ts
+++ b/src/classes/UpdateBehaviourClass.ts
@@ -25,21 +25,24 @@ export default class UpdateBehaviourClass extends PIXI.Container {
   private _intervalFrequency: number;
   private _hoverNode: boolean;
   private _hover: boolean;
+  node: PPNode;
 
   constructor(
     inUpdate: boolean,
     inInterval: boolean,
-    inIntervalFrequency: number
+    inIntervalFrequency: number,
+    node: PPNode,
   ) {
     super();
 
     this._update = inUpdate;
     this._interval = inInterval;
     this._intervalFrequency = inIntervalFrequency;
+    this.node = node;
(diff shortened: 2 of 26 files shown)


src/classes/UpdateBehaviourClass.ts
```diff
@@ -168,7 +168,7 @@ export default class UpdateBehaviourClass extends PIXI.Container {
   // METHODS
 
   getNode(): PPNode {
-    return this.parent as PPNode;
+    return this.parent?.parent as PPNode;
```
magnificus, line 171: could we turn this into something that doesnt potentially return null? Maybe instead of using parent field, add a field on the UpdateBehaviourClass object that is always set on initialization?

src/classes/SocketClass.tsx
```diff
@@ -337,7 +337,7 @@ export default class Socket extends PIXI.Container implements Tooltipable {
   }
 
   getNode(): PPNode {
-    return this.parent as PPNode;
+    return this.parent?.parent as PPNode;
```
magnificus, line 340: same etc

magnificus (approved):
Looks good but would be nice to take the opportunity to get rid of null-potentials in the code

src/classes/UpdateBehaviourClass.ts
fakob: Hmm, I am not sure what you mean. Can you elaborate or take over this part?
Without it we get an error which I believe is because of the `getAllNodeTypes` function where a class instance is created without actually being drawn.

magnificus: sorry for no response, just meant that you check here for null via ".?",  I think it is better to not null check and instead make sure it is never null to begin with

magnificus: if you are not sure when and how this.parent behaves, then you can pass in the node to UpdateBehaviourClass when it is created, and return this.node instead

fakob: I understand that you would like to get rid of the null case :-)
If I would pass in the node on creation, it has to be part of the constructor. As the behaviourClass structure is 
* can be overwritten
* and is stored in the graph file
I would have to pass in the node in all these cases I guess. Or is there another way?

Screenshot 2023-09-22 at 17 53 46

magnificus: I would imagine passing it just as it is instantiated on the node, because that is when it is instantiated right? Would not have to serialize, this is not a massive deal but

fakob: Just so I get it right. 
I add it in the class constructor
Screenshot 2023-09-25 at 18 06 40

And then pass it in wherever I instantiate it
Screenshot 2023-09-25 at 18 07 20

For the SocketClass mentioned below, passing the node in the same way, feels strange.

magnificus: I think this looks good, then we make our own abstraction instead of piggybacking onto the fact that the updatebehaviour is attached visually to the node, this will work even if we spawn the updatebehaviour somewhere else visually, old behaviour would break.

PR #312, opened 2023-09-27
Update graph search header

fakob:
* updated the graph search header
  * now it shows the name of the selected graph
  * added save option
  * moved the share button inside the bar
  * added a create new/empty graph button
  * made it work for small screens
* added share option to context menu

Screenshot 2023-09-27 at 19 38 24
Screenshot 2023-09-27 at 19 38 14

### DIFF
--- a/src/PPStorage.tsx
+++ b/src/PPStorage.tsx
@@ -268,6 +268,17 @@ export default class PPStorage {
     }
   }
 
+  async getGraphNameFromDB(graphId = undefined): Promise<undefined | string> {
+    try {
+      const loadedGraphId = PPGraph.currentGraph.id;
+      const graph = await this.db.graphs.get(graphId || loadedGraphId);
+      return graph.name;
+    } catch (e) {
+      console.log(e.stack || e);
+      return undefined;
+    }
+  }
+
   async getGraphFromDB(id: string): Promise<undefined | Graph> {
     try {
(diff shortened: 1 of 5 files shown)


src/components/Search.tsx
```diff
@@ -21,48 +25,197 @@ import match from 'autosuggest-highlight/match';
...
+              props.setShowSharePlayground(true);
+            }}
+          >
+            Share
+          </Button>
+          {props.isLoggedIn && (
+            <Button
+              variant="text"
+              size="small"
+              title="Logour from Github"
```
magnificus, line 95: small typo

src/PPStorage.tsx
```diff
@@ -268,6 +268,17 @@ export default class PPStorage {
     }
   }
 
+  async getGraphNameFromDB(graphId = undefined): Promise<undefined | string> {
```
magnificus, line 271: This function has a lot of different behaviors, maybe you should be forced to send in graphId and return always string? (if there is no graph found then can return "GRAPH_NAME_NOT_FOUND or something"

magnificus (approved):
cool

src/components/Search.tsx
fakob: Good catch!

src/PPStorage.tsx
fakob: Sounds good

PR #314, opened 2023-10-04
Add help

fakob:
* Added a left side drawer with a create, explore and about section
  * Ctrl+H for showing the help drawer
* Adjusted the general drawer styling a bit and moved the open close icons to the bottom left and right
* Added mdx so we can write the documentation in markdown with react components

Screenshot 2023-10-04 at 22 45 45


magnificus (approved):
cool

PR #315, opened 2023-10-10
Dynamic inputs

magnificus:
gonna add make json and combine graphics to this concept as well (unless you protest)

### DIFF
--- a/src/classes/SocketClass.tsx
+++ b/src/classes/SocketClass.tsx
@@ -523,6 +523,7 @@ export default class Socket extends PIXI.Container implements Tooltipable {
   }
 
   destroy(): void {
+    PPGraph.currentGraph.socketHoverOut(this);
     super.destroy();
     this.getNode()?.resizeAndDraw();
   }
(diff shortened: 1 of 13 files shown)


src/classes/SocketClass.tsx
```diff
@@ -523,7 +523,7 @@ export default class Socket extends PIXI.Container implements Tooltipable {
   }
 
   destroy(): void {
+    PPGraph.currentGraph.socketHoverOut(this);
     super.destroy();
-    this.getNode()?.resizeAndDraw();
```
fakob, line 527: Are we now redrawing somewhere else when a socket is being deleted? 2 cases come to mind. Deleting an added trigger socket. Deleting a socket which existed on the node of the saved graph file, but does not exist on the new node.

fakob (approved):
Does this dynamically add and remove inputs when connecting to it so there is always one available? I think this is a great feature and it would definitely make sense to adapt this to others. Maybe also the math nodes?

src/classes/SocketClass.tsx
magnificus: Yes it dynamically adds when you drop a link on top of it and vice versa, i will check out the case of deleting an added trigger socket, trying to make sure we dont redraw all over the place

magnificus: deleting added trigger socket seems ait to me, I think this will remove a lot of unneccessary draws

magnificus: Adding to add node is a good idea, any other math nodes that would work you think?

PR #316, opened 2023-10-14
Pnp header improvements

fakob:
* moved header bar with search into its separate component/file
* added option to easily rename graphs
* added a save image button to the video example


magnificus (approved):
looks promising

PR #318, opened 2023-10-24
made break unpack

magnificus:


fakob (approved):
The PR looks good, but when testing it crashed when I edited the input json.
1. I added a code editor node and pasted JSON data inside
2. I connected a break node
3. I changed the JSON data so it breaks/can not be parsed
4. PnP crashes

magnificus:
OK will check out before merge

PR #319, opened 2023-10-25
Xml reader

fakob:
* Added an XML reader node which
  * dynamically loads the xml2js package
  * allows conversion from XML to JSON

Screenshot 2023-10-25 at 16 37 20

### DIFF
--- a/src/nodes/data/xml.tsx
+++ b/src/nodes/data/xml.tsx
@@ -1,0 +1,82 @@
+import PPSocket from '../../classes/SocketClass';
+import { CustomArgs, TNodeSource, TRgba } from '../../utils/interfaces';
+import { NODE_TYPE_COLOR, SOCKET_TYPE } from '../../utils/constants';
+import PPNode from '../../classes/NodeClass';
+import { JSONType } from '../datatypes/jsonType';
+import { CodeType } from '../datatypes/codeType';
+import PPGraph from '../../classes/GraphClass';
+
+const outputSocketName = 'output';
+const inputSocketName = 'input';
+
+const IMPORT_NAME = 'xml2js';
+
+export class XMLReader extends PPNode {
+  initialData: any;
+  xml2jsModule;
+  xml;
+
+  constructor(name: string, customArgs?: CustomArgs) {
+    super(name, {
+      ...customArgs,
+    });
+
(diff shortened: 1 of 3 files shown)


src/nodes/data/xml.tsx
```diff
@@ -0,0 +1,78 @@
...
+        outputSocketName,
+        new JSONType(),
+        undefined,
+        true,
+      ),
+      new PPSocket(SOCKET_TYPE.IN, inputSocketName, new CodeType(), '', true),
+    ];
+  }
+
+  public onNodeAdded = async (source?: TNodeSource): Promise<void> => {
```
magnificus, line 56: should not need this, it should automatically get the import if declared as dynamic. Would skip the initial data thing as well

src/nodes/data/xml.tsx
```diff
@@ -0,0 +1,78 @@
...
+    if (this.initialData) {
+      this.setInputData(inputSocketName, this.initialData);
+    }
+    this.executeOptimizedChain();
+
+    super.onNodeAdded(source);
+  };
+
+  protected async onExecute(input, output): Promise<void> {
+    const result = await this.xml2jsModule.parseStringPromise(
```
magnificus, line 68: I would avoid saving this.xml2jsmodule at all, can just use it directly from PPGraph here

src/nodes/data/xml.tsx
magnificus: Would skip this entire function

magnificus (approved):
nice, and nice with dynamic import

src/nodes/data/xml.tsx
fakob: Ah, nice. Didn't know. Will do

src/nodes/data/xml.tsx
fakob: I can remove the import line, but the initial data I need. Otherwise I can not get the data when I drag in the XML file. Is there another way of doing this?
I also needed the execute line as otherwise the output would not show anything. Also here, is there another way?

fakob: Ah, I think I can use the 
defaultArguments: {
    [socketName]: initialData,
}

Forgot about that

PR #321, opened 2023-10-27
Fix event from within text input issue and small cleanup

fakob:
When using Shift+2 or Shift 1 inside the label, it would detect the zoom commands instead of writing @ or !
I noticed that we had two functions for it
* isEventComingFromWithinTextInput
* isEventComongFromWithinWidget

I have now consolidated them into 1. Not sure what the reason was why we/I had thought that there has to be a difference. Do you know?


magnificus (approved):
nice

PR #328, opened 2023-11-03
Add individual urls

fakob:
Added a URL parameter (`loadLocal`) when a local graph is loaded. This allows you to
* load different graphs in different tabs
* go back and forth between different graphs
If there is no `loadLocal` parameter, we fall back to loading the last saved graph.

https://github.com/fakob/plug-and-play/assets/4619772/8be456b8-588f-474c-b756-0fef065c3b45

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -262,7 +262,8 @@ Viewport position (scale): ${viewportScreenX}, ${Math.round(
     );
 
     window.addEventListener('popstate', () => {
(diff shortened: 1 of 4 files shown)


magnificus (approved):
nice that its moved away from App.tsx, but could we make loadGraph() take the input as parameter instead of reading the source itself? beomes more modular

fakob:
Sounds good.

PR #330, opened 2023-11-04
graph parse incoming data aggressively

magnificus:


magnificus:
The nice thing about this is that you can now much more often just drag in an excel sheet and then directly link to a graph it will try to parse it so it doesnt have to be an exact match, planning to add more stuff for this

fakob (approved):

PR #332, opened 2023-11-08
First version of format JSON, missing a lot of polish and QOL

magnificus:


fakob (approved):
Looks cool!
The only input I have is about the node names and descriptions. The node names I would have chosen in a way that they maybe start with the same word so they are ordered along side of each other?
* Format object properties
* Format object properties (map)

Regarding the description I would have maybe used the word customise or transform to heighten the chances that it will be found when searched?

magnificus:
OK will change names and add those transforms, im hoping to add some more polish to these

PR #333, opened 2023-11-09
only certain options in dropdown

magnificus:


fakob (approved):
Looks good.

With the Filetype, I am a bit unsure as it outputs a string (this might change in the future). There might be the corner case of using this in a custom function. This would be an advanced handling though.
Would it be possible to hide the overflow menu and disable the button, in case the type is not part of the list?

Screenshot 2023-11-09 at 21 26 13

fakob:
But I guess the same advanced programming concern would be valid for the other types?

magnificus:
yes, an advanced menu could make sense

magnificus:
With filetype, does that mean you would like it part of the dropdown options?

PR #337, opened 2023-11-11
Dashboard and left side menu

fakob:
* Added dashboard
* Updated left side menu to include graphs list
* Removed pnp header
* Updated shortcuts for toggling left side menu `1`, dashboard `2`, and right side menu `3`
* Added rename test
* Updated the focus layer to ignore mouse interaction so it does not block dragging/selecting
* Updated toggle functions to optionally force open/close

Screenshot 2023-11-12 at 12 17 08
Screenshot 2023-11-12 at 12 16 42
Screenshot 2023-11-12 at 12 16 26


src/classes/GraphClass.ts
```diff
@@ -1114,6 +1125,10 @@ export default class PPGraph {
       console.log(error);
       return false;
     }
+
+    // create layouts
+    this.layouts = data?.layouts ? data.layouts : LAYOUTS_EMPTY;
```
magnificus, line 1130: this.layouts = data?.layouts || LAYOUTS_EMPTY;

src/classes/SocketClass.tsx
```diff
@@ -406,6 +414,19 @@ export default class Socket extends PIXI.Container implements Tooltipable {
   }
 
   getTooltipContent(props): React.ReactElement {
+    console.log(props);
```
magnificus, line 417: remove log

src/components/GraphOverlayDashboard.tsx
```diff
@@ -0,0 +1,359 @@
...
+      );
+      console.log(newCurrentLayout);
+      setCurrentLayout(newCurrentLayout);
+    }
+  }, [PPGraph.currentGraph?.layouts]);
+
+  const addToDashboard = (socket: PPSocket) => {
+    InterfaceController.toggleShowDashboard(true);
+    setCurrentLayout((prevLayout) => {
+      console.log('addToDashboard', socket);
```
magnificus, line 75: some logs around here that can be removed

src/components/GraphOverlayDrawer.tsx
```diff
@@ -13,28 +17,44 @@ type GraphOverlayDrawerProps = {
...
-  const [helpDrawerWidth, setHelpDrawerWidth] = useState(
-    defaultHelpDrawerWidth,
-  );
-  const [drawerWidth, setDrawerWidth] = useState(defaultDrawerWidth);
+  const [leftDrawerWidth, setLeftDrawerWidth] = useState(DEFAULT_DRAWER_WIDTH);
+  const [rightDrawerWidth, setRightDrawerWidth] =
+    useState(DEFAULT_DRAWER_WIDTH);
+
+  useEffect(() => {
+    console.log('onDrawerSizeChanged');
```
magnificus, line 25: here as well etc

src/components/ResponsiveDrawer.tsx
```diff
@@ -42,53 +45,63 @@ function DrawerToggleInspector(props) {
...
     </Box>
   );
 }
 
 const ResponsiveDrawer = (props) => {
   // leaving this commented here for potential future testing
-  const [open, setOpen] = useState(true);
-  const [filter, setFilter] = useState(null);
-  const [helpFilter, setHelpFilter] = useState('create');
+  const [nodeFilter, setNodeFilter] = useState(null);
```
magnificus, line 91: instead of null, could the default filter not be () => true ?

src/components/ResponsiveDrawer.tsx
```diff
@@ -42,53 +45,63 @@ function DrawerToggleInspector(props) {
...
-  const toggleInspectorAndFocus = ({ filter, socket, open }) => {
+  const toggleInspectorAndFocus = ({ filter, socket }) => {
+    InterfaceController.toggleRightSideDrawer(true);
     if (!props.isLeft) {
-      if (open !== undefined) {
-        setOpen(open);
-      } else {
-        handleDrawerToggle();
-      }
       if (filter) {
```
magnificus, line 103: then also dont need to check it here

src/nodes/datatypes/abstractType.tsx
```diff
@@ -48,6 +56,14 @@ export class AbstractType {
     return <DefaultOutputWidget {...props} />;
   };
 
+  getInputWidgetSize(): any {
```
magnificus, line 59: avoid copy pasting 

  getInputWidgetSize(): any {
    return widgetSize;
  }
  
    getOutputWidgetSize(): any {
    return widgetSize;
  }
  
  
  in every class. Instead define a 
  
  getDefaultWidgetSize(){
  return etc:
  }
  
  that you override in all classes, then just in abstractType, make
  
 getOutputWidgetSize(): any {
    return getDefaultWidgetSize();
  }
  
   getInputWidgetSize(): any {
    return getDefaultWidgetSize();
  }
  
  then you dont have to copy paste these two functions in all child classes, just override default, and if you want to have different for input and output, override them separately.

magnificus (changes requested):
Very cool, looks good apart from the copy pasting in dataTypes, everything else nice

src/nodes/datatypes/abstractType.tsx
fakob: Ah, yes. Good point.

src/components/ResponsiveDrawer.tsx
fakob: null is used to represent no filter, this means that all content is shown. I can change this to an empty string if you thing that is better. then I have to check if it is an empty string later though.

magnificus (approved):

src/components/ResponsiveDrawer.tsx
magnificus: What you could do instead is to have default be a filter that returns allows everything through, could be cleaner

fakob: I tried that now, but forgot that I am using the https://mui.com/material-ui/api/toggle-button-group/ which works with null when nothing is selected/filtered. Will leave it for now.

PR #341, opened 2023-11-19
Fix for header switching to string on http node

fakob:
I suggest to only adapt the output?


magnificus (approved):
OK good idea. Could use just

return !socket.isInput()

PR #342, opened 2023-11-19
Update parsing with warning

fakob:
This is the first step of updating parsing.
* now parsing a value always returns a valid value of the specific data type
* in case there are errors in the parsing, we include a warning
  * for now I have used only a single string at the moment, but can be made into an array in the future
* currently it only triggers a warning log. In the second part I will trigger an Error from the Error class and show this warning somewhere. This is the second step

### DIFF
--- a/src/InterfaceController.ts
+++ b/src/InterfaceController.ts
@@ -22,8 +22,6 @@ export enum ListenEvent {
   GlobalPointerUpAndUpOutside, // data = event: PIXI.FederatedPointerEvent
   GraphChanged, // data = {id,name}
   ToggleInspectorWithFocus, // (data: Node, Filter or Socket) => void: called when inspector should be opened with focus
-  onNodeStatusChanged, // data = node
-  onSocketStatusChanged, // data = socketId
   EscapeKeyUsed,
   UnsavedChanges, // data = Boolean
   ResourceUpdated, // data = {id}
@@ -44,8 +42,6 @@ export default class InterfaceController {
     10: {},
     11: {},
     12: {},
-    13: {},
-    14: {},
   }; // not sure why this one is so messed up and needs these defined by default, very annoying
 
   // we use this listener structure here as there can be multiple listeners, not needed for everything (sometimes there is just one listener)
--- a/src/InterfaceController.ts
+++ b/src/InterfaceController.ts
@@ -22,6 +22,8 @@ export enum ListenEvent {
   GlobalPointerUpAndUpOutside, // data = event: PIXI.FederatedPointerEvent
   GraphChanged, // data = {id,name}
   ToggleInspectorWithFocus, // (data: Node, Filter or Socket) => void: called when inspector should be opened with focus
+  onNodeStatusChanged, // data = node
+  onSocketStatusChanged, // data = socketId
   EscapeKeyUsed,
   UnsavedChanges, // data = Boolean
   ResourceUpdated, // data = {id}
@@ -42,6 +44,8 @@ export default class InterfaceController {
     10: {},
     11: {},
     12: {},
+    13: {},
+    14: {},
   }; // not sure why this one is so messed up and needs these defined by default, very annoying
 
   // we use this listener structure here as there can be multiple listeners, not needed for everything (sometimes there is just one listener)
--- a/src/NodeArrayContainer.tsx
+++ b/src/NodeArrayContainer.tsx
@@ -107,9 +107,9 @@ const NodeItem = (props) => {
   );
 
   useInterval(() => {
-    const newNnodeStatus = props.property.status.node as PNPStatus;
+    const newNodeStatus = props.property.status.node as PNPStatus;
     const newSocketStatus = props.property.status.node as PNPStatus;
-    if (socketStatus !== newSocketStatus || nodeStatus !== newNnodeStatus) {
+    if (socketStatus !== newSocketStatus || nodeStatus !== newNodeStatus) {
       setNodeStatus(props.property.status.node as PNPStatus);
       setSocketStatus(props.property.status.socket as PNPStatus);
     }
--- a/src/NodeArrayContainer.tsx
+++ b/src/NodeArrayContainer.tsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useEffect, useState } from 'react';
+import useInterval from 'use-interval';
 import {
   Box,
   Button,
@@ -98,8 +99,21 @@ const NodesContent = (props) => {
 };
 
 const NodeItem = (props) => {
-  const nodeStatus = props.property.status.node as PNPStatus;
-  const socketStatus = props.property.status.socket as PNPStatus;
+  const [nodeStatus, setNodeStatus] = useState(
+    props.property.status.node as PNPStatus,
+  );
+  const [socketStatus, setSocketStatus] = useState(
+    props.property.status.socket as PNPStatus,
+  );
+
+  useInterval(() => {
+    const newNnodeStatus = props.property.status.node as PNPStatus;
+    const newSocketStatus = props.property.status.node as PNPStatus;
+    if (socketStatus !== newSocketStatus || nodeStatus !== newNnodeStatus) {
+      setNodeStatus(props.property.status.node as PNPStatus);
+      setSocketStatus(props.property.status.socket as PNPStatus);
+    }
+  }, 100);
 
   return (
     <ListItem
@@ -483,9 +497,6 @@ export const NodeArrayContainer: React.FunctionComponent<
     ids.push(
       InterfaceController.addListener(ListenEvent.GraphChanged, () => {
         updateNodesAndInfo();
-      }),
-      InterfaceController.addListener(ListenEvent.onNodeStatusChanged, () => {
-        updateNodesAndInfo(false);
       }),
     );
 
--- a/src/NodeArrayContainer.tsx
+++ b/src/NodeArrayContainer.tsx
@@ -21,16 +21,13 @@ import styles from './utils/style.module.css';
 import InterfaceController, { ListenEvent } from './InterfaceController';
 import { getConfigData, writeTextToClipboard } from './utils/utils';
 import { ensureVisible, zoomToFitNodes } from './pixi/utils-pixi';
-import {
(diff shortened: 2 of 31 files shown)


src/nodes/datatypes/arrayType.tsx
```diff
@@ -32,15 +32,33 @@ export class ArrayType extends AbstractType {
...
-        return data;
+        warning = warn;
+        parsedData = [];
       }
+    } else {
+      warning = warn;
+      parsedData = [];
     }
-    return data;
+
```
magnificus, line 57: instead of all these lines, at the end, do:

if (parsedData == undefined){
 warnings.push(new NodeTypeWarning('Not an array. [] is returned'))
 parsedData = []

}

src/nodes/datatypes/abstractType.tsx
```diff
@@ -81,8 +83,8 @@ export class AbstractType {
     return TRgba.fromString(SOCKET_COLOR_HEX);
   }
 
-  parse(data: any): any {
-    return data;
+  parse(data: any): TParseType {
```
magnificus, line 86: good with a type

src/utils/interfaces.tsx
```diff
@@ -119,6 +119,11 @@ export type TSocketId =
 
 export type TSocketType = (typeof SOCKET_TYPE)[keyof typeof SOCKET_TYPE];
 
+export type TParseType = {
```
magnificus, line 122: instead of this I would make it

{
value : any;
warnings[]: PNPWarningClass
}

This gives us two things, first, an array both allows more than one warning at a time, and avoids using null. The warning class allows us to add more behaviours and aligns with our error classes.

src/nodes/datatypes/colorType.tsx
```diff
@@ -28,13 +29,30 @@ export class ColorType extends AbstractType {
     return TRgba.randomColor(); //TRgba.fromString('#ff3700');
   }
 
-  parse(data: any): any {
+  parse(data: any): TParseType {
+    let parsedData;
+    let warning: string;
+    const warn = 'Not a color. Default color is returned';
+
     if (typeof data === 'string') {
```
magnificus, line 37: similar to above, can achieve desired result with fewer lines

src/nodes/datatypes/stringType.tsx
```diff
@@ -58,3 +55,28 @@ export class StringType extends AbstractType {
     return ['Label', 'Constant', 'TextEditor'];
   }
 }
+
+export const parseString = (data: any): TParseType => {
+  let parsedData;
+  let warning: string;
+  const warn = 'Not a file type. Empty string is returned';
+
+  if (data === null || data === undefined) {
```
magnificus, line 64: same as further up, can use fewer lines

src/nodes/draw/draw.tsx
```diff
@@ -680,9 +683,10 @@ export class DRAW_Line extends DRAW_Base {
       ],
     };
     const graphics: PIXI.Graphics = new PIXI.Graphics();
-    const selectedColor: TRgba = new ColorType().parse(
+    const { value: selectedColor, warning } = new ColorType().parse(
```
magnificus, line 686: these lines in the draw nodes look copy pasted, turn into common function

magnificus (changes requested):
looks promising and good conceptually but have some input on code

src/utils/interfaces.tsx
fakob: Sounds good. I wanted to do this in the seconds step. But can also do it now.

src/nodes/datatypes/arrayType.tsx
fakob: Good idea

src/nodes/datatypes/colorType.tsx
fakob: will do

fakob:
There are 2 known limitations
* If a node is not updated, the warning only disappears if it is executed, or selected. Selecting triggers the PropertyArrayContainer > getConfigData > serialize > get data and is therefore updated/reparsed
* The errors/warnings in the node browser do not update unless you click in the graph to send unselect

src/SocketContainer.tsx
```diff
@@ -34,6 +37,7 @@ export const SocketContainer: React.FunctionComponent<SocketContainerProps> = (
   props,
 ) => {
   const myRef = useRef(null);
+  const [hasError, setHasError] = useState(props.property.status.isError());
```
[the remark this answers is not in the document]
magnificus: not a massive deal, points below are more important, you decide on this

src/classes/SocketClass.tsx
```diff
@@ -119,53 +125,79 @@ export default class Socket extends PIXI.Container implements Tooltipable {
...
-      this.data,
+      this._data,
     );
     this._ValueSpecificGraphics.x = this.getSocketLocation().x;
     this._ValueSpecificGraphics.y = this.getSocketLocation().y;
     this.addChild(this._ValueSpecificGraphics);
   }
 
+  public setStatus(status: PNPStatus) {
+    const currentMessage = JSON.stringify(this.status.message);
```
magnificus, line 148: isnt message always a string? if so dont need to stringify

src/nodes/api/chatgpt.tsx
```diff
@@ -77,12 +78,11 @@ export class ChatGPTNode extends HTTPNode {
     inputObject: any,
     outputObject: Record<string, unknown>,
   ): Promise<void> {
-    this.statuses = [];
+    this.status.custom = [];
```
magnificus, line 81: it seems like most of these only want one of these custom notifications at a time right? If so we could have a function on NodeClass "pushExclusiveCustomStatus", then we wouldnt need to set this.status.custom = [] manually everywhere we do this,

src/InterfaceController.ts
```diff
@@ -22,6 +22,8 @@ export enum ListenEvent {
   GlobalPointerUpAndUpOutside, // data = event: PIXI.FederatedPointerEvent
   GraphChanged, // data = {id,name}
   ToggleInspectorWithFocus, // (data: Node, Filter or Socket) => void: called when inspector should be opened with focus
+  onNodeStatusChanged, // data = node
```
magnificus, line 25: might be easier to skip these, left comment further down

magnificus (changes requested):
Looks very promising, left some comments

src/NodeArrayContainer.tsx
```diff
@@ -449,20 +456,26 @@ export const NodeArrayContainer: React.FunctionComponent<
 
   const customSort = (a: PPNode, b: PPNode) => {
     const order =
-      (+b.status.isError() - +a.status.isError()) * 100 +
+      (+b.status.node.isError() - +a.status.node.isError()) * 1000 +
```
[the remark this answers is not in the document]
fakob: Using the + here is the easiest way to convert the boolean to a number for typescript. Or should I use a different way?

src/classes/NodeClass.tsx
```diff
@@ -722,33 +731,48 @@ export default class PPNode extends PIXI.Container {
     this._StatusesRef.clear();
     this._StatusesRef.removeChildren();
 
-    this.statuses.forEach((nStatus, index) => {
-      const color = nStatus.color;
+    let flattenedStatus = [];
+    for (const key in this.status) {
+      if (Array.isArray(this.status[key])) {
```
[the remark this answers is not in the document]
fakob: The `status` on the node itself is not a class, only the items inside. I am using this to get flatten this object/array combination. I can also do something like this:

const flattened = [];
this.status.node.isError && flattened.push(this.status.node);
this.status.socket.isError && flattened.push(this.status.socket);
flattened.push(...this.status.custom);

Is this better?

Screenshot 2023-12-05 at 18 24 49

src/classes/NodeClass.tsx
```diff
@@ -810,12 +832,34 @@ export default class PPNode extends PIXI.Container {
...
+      this.drawStatuses();
+      this.drawErrorBoundary();
+      InterfaceController.notifyListeners(
+        ListenEvent.onNodeStatusChanged,
+        this,
+      );
+    }
+  }
+
+  doSocketsHaveErrors(): void {
```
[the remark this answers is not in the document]
fakob: Good idea

src/classes/NodeClass.tsx
```diff
@@ -810,12 +832,34 @@ export default class PPNode extends PIXI.Container {
     });
   }
 
-  protected setStatus(status: PNPStatus) {
-    if (
-      JSON.stringify(this.status.message) !== JSON.stringify(status.message)
-    ) {
-      this.status = status;
-      this.drawNodeShape();
+  public setStatus(status: PNPStatus, typeSocket = false) {
```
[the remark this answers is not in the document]
fakob: Good idea

src/utils/utils.ts
```diff
@@ -903,3 +910,31 @@ export const loadGraph = (urlParams: URLSearchParams) => {
 
 export const sortByDate = (a, b) =>
   new Date(b.date).getTime() - new Date(a.date).getTime();
+
+export const parseValueAndAttachWarnings = (
```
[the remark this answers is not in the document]
fakob: Will do

src/NodeArrayContainer.tsx
```diff
@@ -101,6 +99,22 @@ const NodesContent = (props) => {
...
 const NodeItem = (props) => {
+  const [nodeStatus, setNodeStatus] = useState(
+    props.property.status.node as PNPStatus,
+  );
+  const [socketStatus, setSocketStatus] = useState(
+    props.property.status.socket as PNPStatus,
+  );
+
+  useInterval(() => {
+    const newNnodeStatus = props.property.status.node as PNPStatus;
```
magnificus, line 110: very minor, but typo on newNodeStatus

magnificus (approved):
Very cool, looks very useful

src/NodeArrayContainer.tsx
fakob: Wow, good catch!

PR #344, opened 2023-11-20
Constructor demotion 2

magnificus:
This is a lower energy version of the constructor demotion, but it is safer and should give the same (?) performance increase, there is definitely room for further improvements though


fakob (approved):
Looks good to me, however, could you please check
* the "Using macros ..." examples. They stopped working with this version
* also the image node has an issue. Add an image node, resize it and save the file. Then load it again. The image is not resized to the previous size.
This was not a thorough check, but just clicking through the examples.

PR #348, opened 2023-12-06
allow header injection for html node

fakob:
* added a header socket which allows to add code in the header of the html node
  * as default example we are loading tailwind and style the text with it

Screenshot 2023-12-06 at 22 35 46

### DIFF
--- a/src/nodes/draw/html.tsx
+++ b/src/nodes/draw/html.tsx
@@ -18,15 +18,6 @@ import HybridNode2 from '../../classes/HybridNode2';
 const inputSocketNameHeader = 'Header';
 const inputSocketNameHtml = 'Html';
 const reloadSocketName = 'Reload';
-const defaultHeader = '<script src="https://cdn.tailwindcss.com"></script>';
-const defaultCode = `<div class="p-4">
-<h2>HTML Node</h2>
-<p class="mb-2 text-sky-500 dark:text-sky-400">Embed an iframe or write your own HTML</p>
-<form>
-  <button class="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded" formtarget="_blank" formaction="https://github.com/fakob/plug-and-play/">Click me!</button>
-</form>
-</div>
-`;
 
 export class HtmlRenderer extends HybridNode2 {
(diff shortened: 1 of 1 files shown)


src/nodes/draw/html.tsx
```diff
@@ -101,47 +111,59 @@ export class HtmlRenderer extends HybridNode2 {
 
   // small presentational component
   protected getParentComponent(props: any): React.ReactElement {
-    const node = props.node;
-    const iframeRef = useRef();
-    const [htmlData, setHtmlData] = useState(props[inputSocketName]);
-    const [reload, setReload] = useState(props[reloadSocketName]);
-
-    useEffect(() => {
-      if (iframeRef.current) {
```
magnificus, line 110: when would this be invalid? can check be removed?

src/nodes/draw/html.tsx
```diff
@@ -190,7 +212,7 @@ export class EmbedWebsite extends HtmlRenderer {
   protected getDefaultIO(): PPSocket[] {
     const codeSocket = super
       .getDefaultIO()
-      .find((socket) => socket.name === inputSocketName);
+      .find((socket) => socket.name === inputSocketNameHtml);
```
magnificus, line 215: You can do this much more elegantly with inheritance, change defaultCode in the parent to

getDefaultHTMLCode() (or some other name), then in parent getDefaultIO

      new PPSocket(
        SOCKET_TYPE.IN,
        inputSocketNameHtml,
        new CodeType(),
        getDefaultHTMLCode(),
        false,

then in child just override

getDefaultHTMLCode() with
return `<iframe src="https://en.wikipedia.org/wiki/Special:Random" style="width: 100%; height: 100%;"></iframe>`

magnificus (changes requested):

src/nodes/draw/html.tsx
fakob: Good idea!

src/nodes/draw/html.tsx
fakob: Good question. I think with the constructor update this is no longer an issue. Will remove.

magnificus (approved):

PR #349, opened 2023-12-07
Add update on load

fakob:
* Added an "Update on load" option which executes the node after everything has been loaded
  * This is helpful for nodes like the HTTP node or Custom functions which should be executed once the graph is loaded, but which you might not want to update always on change

This is different from "Execute on place" which is executed when nothing is connected. It makes sense to keep this hidden away from the user as some nodes rely on it.

Screenshot 2023-12-07 at 22 30 30

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -1115,7 +1115,10 @@ export default class PPGraph {
       ),
     );
     await FlowLogic.executeOptimizedChainBatch(
-      nodes.filter((node) => node.updateBehaviour.load),
+      nodes.filter((node) => {
+        console.log(node.name);
+        return node.updateBehaviour.load;
+      }),
     );
   }
 
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -1114,6 +1114,9 @@ export default class PPGraph {
         (node) => !node.getHasDependencies() && node.updateBehaviour.update,
       ),
     );
+    await FlowLogic.executeOptimizedChainBatch(
+      nodes.filter((node) => node.updateBehaviour.load),
+    );
   }
 
   getInputSocket(nodeID: string, socketName: string): PPSocket {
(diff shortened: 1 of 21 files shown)


fakob:
> This is different from "Execute on place" which is executed when nothing is connected. It makes sense to keep this hidden away from the user as some nodes rely on it.

As execute on place was, I guess, mainly fixing specific nodes which need to be executed once. This can probably, after your refactor, be done in the onNodeAdded function. Then we can get rid of execute on place I guess.

src/classes/NodeClass.tsx
```diff
@@ -159,13 +159,11 @@ export default class PPNode extends PIXI.Container implements IWarningHandler {
...
-    this.getAllSockets().forEach((socket) =>
-    {
+    this.getAllSockets().forEach((socket) => {
       this._BackgroundRef.addChild(socket);
       socket.onNodeAdded();
-    }
-    );
-    if (this.executeOnPlace() || this.updateBehaviour.update) {
+    });
+    if (this.executeOnPlace()) {
```
magnificus, line 166: can we get rid of executeonplace and instead use load now? :), alternatively, just make executeonplace return this value. But we also want to remove executeOnPlace in all children, and instead make them use the new load value

magnificus (changes requested):
looks good, now we just want to get rid of remains from old system

src/classes/NodeClass.tsx
fakob: Will do

src/classes/GraphClass.ts
```diff
@@ -1114,6 +1114,12 @@ export default class PPGraph {
         (node) => !node.getHasDependencies() && node.updateBehaviour.update,
       ),
     );
+    await FlowLogic.executeOptimizedChainBatch(
+      nodes.filter((node) => {
+        console.log(node.name);
```
magnificus: can remove log

fakob: just did :-)

magnificus: BTW, you can probably combine this with the above one, so that we dont have to execute a potential node that is both seed node and load two times

magnificus (approved):
looks good now just one thing about potentially having to execute some initial nodes twice, not neccessary

src/classes/GraphClass.ts
fakob: I will create a separate pull request for this.

PR #351, opened 2023-12-09
Significant performance improvements when dealing with larger data

magnificus:
…ts called in interface, sped it up

there are more things I found I will fix

### DIFF
--- a/src/nodes/datatypes/jsonType.tsx
+++ b/src/nodes/datatypes/jsonType.tsx
@@ -3,7 +3,6 @@ import React from 'react';
 import { SocketParsingWarning } from '../../classes/ErrorClass';
 import { JSONWidget } from '../../widgets';
 import { AbstractType, DataTypeProps } from './abstractType';
-import { convertToString } from '../../utils/utils';
 import { TParseType, TRgba } from '../../utils/interfaces';
 
 export interface JSONTypeProps extends DataTypeProps {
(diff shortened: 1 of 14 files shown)


src/nodes/datatypes/jsonType.tsx
```diff
@@ -23,9 +22,6 @@ export class JSONType extends AbstractType {
 
   getInputWidget = (props: JSONTypeProps): any => {
     props.dataType = this;
-    if (typeof props.property.data !== 'string') {
```
magnificus, line 26: this was setting the data on the socket itself to a stringified version on widget open, was costing a lot of performance and seemed very strange to me

fakob (changes requested):
The point with the "Info" tab for both graph and node is that you can
* see its current config (serialized view)
* edit and replace it

I understand that it is not practical to serialize the graph or node every time you open the inspector, but then I suggest that we introduce a "load config" button in the "Info" tab, which would serialize the graph or node. Currently the config part of the info tab has lost its meaning as it does not show the current state.

Screenshot 2023-12-11 at 21 34 30

magnificus:
OK, load config button sounds good to me, would you be able to help me with that? will then remove the parameter to serialize "includeSockets"

fakob:
I have now pushed a change where the sourceContent is broken out into its own component and in there I have added a load data button. Please have a look.

Screenshot 2023-12-13 at 23 33 42

PR #352, opened 2023-12-13
Update executeAllSeedNodes logic

fakob:
Is it this what we want for executeAllSeedNodes?
* execute all nodes which
  * do not have any dependencies
  * AND have on load enabled

### DIFF
--- a/src/classes/GraphClass.ts
+++ b/src/classes/GraphClass.ts
@@ -1111,7 +1111,9 @@ export default class PPGraph {
   async executeAllSeedNodes(nodes: PPNode[]): Promise<void> {
     await FlowLogic.executeOptimizedChainBatch(
       nodes.filter(
-        (node) => !node.getHasDependencies() && node.updateBehaviour.load,
+        (node) =>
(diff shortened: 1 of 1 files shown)


src/classes/GraphClass.ts
```diff
@@ -1111,14 +1111,9 @@ export default class PPGraph {
   async executeAllSeedNodes(nodes: PPNode[]): Promise<void> {
     await FlowLogic.executeOptimizedChainBatch(
       nodes.filter(
-        (node) => !node.getHasDependencies() && node.updateBehaviour.update,
+        (node) => !node.getHasDependencies() && node.updateBehaviour.load,
```
magnificus, line 1114: might be misunderstanding, but shouldnt it be:

(!node.getHasDependencies() && node.updateBehaviour.update) || node.updateBehaviour.load ?

with the way it is now it wont execute if it is a node that has load checked but say a constant in

fakob: Ah yes, I made a mistake.

fakob: Good catch!

PR #353, opened 2023-12-14
Fixed trigger disappearing from custom function node

fakob:
The custom function node was using `getAllInputSockets` to `adaptInputs` which would remove also the manually added sockets.

### DIFF
--- a/src/nodes/data/dataFunctions.tsx
+++ b/src/nodes/data/dataFunctions.tsx
@@ -288,22 +288,21 @@ export class CustomFunction extends PPNode {
   protected getDefaultParameterValues(): Record<string, any> {
     return {};
   }
+
   protected getDefaultParameterTypes(): Record<string, any> {
     return {};
   }
+
   protected getOutputParameterType(): AbstractType {
     return new AnyType();
   }
+
   protected getOutputParameterName(): string {
     return outDataName;
   }
+
   protected getOutputCodeVisibleByDefault(): boolean {
     return false;
-  }
-  public getAllNonDefaultSockets(): Socket[] {
-    return this.inputSocketArray.filter(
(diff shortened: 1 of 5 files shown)


src/nodes/data/dataFunctions.tsx
```diff
@@ -301,7 +301,7 @@ export class CustomFunction extends PPNode {
     return false;
   }
   public getAllNonDefaultSockets(): Socket[] {
-    return this.getAllInputSockets().filter(
```
magnificus, line 304: This bug you found might be causing problems for others too, we might want to change the definition in NodeClass instead and make it always only care about inputs.

so changing (In NodeClass)

  // get all sockets that are not part of the base kit for the node
  public getAllNonDefaultSockets(): Socket[] {
    const defaultIONames = this.getAllInitialSockets()
      .filter((socket) => socket.isInput())
      .map((socket) => socket.name);
    return this.getAllInputSockets().filter(
      (socket) => !defaultIONames.includes(socket.name),
    );
  }
  
  to
  
    // get all sockets that are not part of the base kit for the node
  public getAllNonDefaultSockets(): Socket[] {
    const defaultIONames = this.getAllInitialSockets()
      .filter((socket) => socket.socketType == SocketType.IN)
      .map((socket) => socket.name);
    return this.getAllInputSockets().filter(
      (socket) => !defaultIONames.includes(socket.name),
    );
  }

fakob: I understand what you mean, but I think there was a small issue in your suggestion as it still removed the triggers.
Please have a look at my update in NodeClass

magnificus: ah yes, looks good

magnificus (approved):

PR #354, opened 2023-12-20
testcontroller + first test

magnificus:


src/TestController.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  getNodeByType(nodeType: string): PPNode {
+    const nodes = Object.values(PPGraph.currentGraph.nodes);
+    return nodes.find((node) => node.type == nodeType);
+  }
+  getNodeCenterByType(nodeType: string): [number, number] {
+    const toReturn = this.getNodeByType(nodeType);
+    return [toReturn.x + toReturn.width / 2, toReturn.y + toReturn.height / 2];
+  }
```
fakob, line 22: I would have thought that the x and y node value is in canvas space and not in screen space, but maybe it works as we have not zoomed or panned? Maybe I misunderstood?

src/TestController.ts
```diff
@@ -0,0 +1,61 @@
...
+import PPLink from './classes/LinkClass';
+import PPNode from './classes/NodeClass';
+
+export default class TestController {
+  identify(): string {
+    return 'its testcontroller';
+  }
+
+  addNode(nodeName): boolean {
+    PPGraph.currentGraph.addNewNode(nodeName);
```
fakob, line 11: Should we return the node Id and add a function to getNodeById? as gettingNodeByType can be ambiguous. Well I guess we can plan our tests to not have nodes of the same type as a start.

fakob (approved):
Very cool that this works. Thanks!

src/TestController.ts
magnificus: It does seem to work when I am trying it, but maybe has some gotcha? not sure

src/TestController.ts
magnificus: yes... I have been working around some annoying limitations, I am currently not able to await any async functions, so I don't know the nodes ID's, but realize we can add a testcontroller function that adds new node with specific ID, then can use.

PR #356, opened 2024-01-14
Fix parsing warning on output

fakob:
Fixes this bug - https://trello.com/c/146AVuwW/99-unexpected-warning-about-output-array-on-multiply-draw-node-when-clicking-one-of-the-outputs

* Fixed outputMultiplierInjected error
* Fixed outputs default values getting overridden on load
* Added test


cypress/e2e/pnp/testMultiplyObject.cy.ts
```diff
@@ -0,0 +1,50 @@
+import { doWithTestController } from './helpers';
```
magnificus, line 1: good with test

magnificus (approved):

PR #358, opened 2024-01-15
removed caching logic from UI

magnificus:


magnificus:
does not fix the big performance thing, posted info about that one in discord

fakob (approved):

PR #359, opened 2024-01-17
Update populating defaults feature

fakob:


src/classes/GraphClass.ts
```diff
@@ -686,6 +691,104 @@ export default class PPGraph {
     this.selectedSourceSocket = null;
   }
 
+  action_AddOrReplaceNode = async (event, selected: INodeSearch) => {
```
magnificus, line 694: I guess this was the same before but we really should split this one up into smaller pieces

src/classes/GraphClass.ts
```diff
@@ -700,12 +803,16 @@ export default class PPGraph {
           overrideId: referenceID,
           nodePosX: node.x + (socket.isInput() ? 0 : node.width + 40),
           nodePosY: node.y + socket.y,
-          initialData: socket.isInput() ? socket.data : undefined,
         },
         NODE_SOURCE.NEWCONNECTED,
       );
       socket.isInput() && newNode.setPosition(-(newNode.width + 40), 0, true);
-      connectNodeToSocket(socket, newNode);
+      connectNodeToSocket(socket, newNode).then(() => {
```
magnificus, line 810: instead of then, i would await. as the caller function is async, so that the caller of that one can in turn await action_addWIdgetNode

magnificus (approved):

src/classes/GraphClass.ts
fakob: Done

PR #360, opened 2024-01-18
Tiny getGraph improvement

fakob:
In Dexie.js there is no easy way of returning just some values. https://github.com/dexie/Dexie.js/issues/468
If this becomes an issue, we can look into some of the suggested solutions.

* I have removed the double sort


magnificus:
That is normal for databases, if we want to fetch these attributes without getting the entire blob we need to store them differently, not on the DB blob object but as separate parameters in the table, tell me if you want help with this, ATM this costs a lot of performance for me so I do think it is a problem

magnificus (approved):

PR #362, opened 2024-01-21
Fix creation errors of some nodes

fakob:
* Fixed error on creation of SqliteReader, Image, JSONSet, Table_GetColumnByName
* Added warning class
* Updated tests


magnificus (approved):
good stuff

PR #363, opened 2024-01-22
Fixed bug with dropdown node when toggling multi selection

fakob:
Fixes this issue - https://trello.com/c/DpnYhJBz

### DIFF
--- a/cypress/e2e/pnp/testWidgetNodes.cy.ts
+++ b/cypress/e2e/pnp/testWidgetNodes.cy.ts
@@ -1,0 +1,68 @@
+import { doWithTestController } from './helpers';
+
+describe('testWidgetNodes', () => {
+  it('testDropdownWidget', () => {
+    cy.visit('http://127.0.0.1:8080/?new=true');
+    cy.wait(100);
+    doWithTestController((testController) => {
+      expect(testController.addNode('Constant', 'Constant')).to.eq(true);
+      expect(testController.addNode('WidgetDropdown', 'WidgetDropdown')).to.eq(
+        true,
+      );
+      expect(testController.addNode('Label', 'Label')).to.eq(true);
+    });
+    cy.get('canvas').click();
+    cy.wait(100);
+    doWithTestController((testController) => {
+      testController.moveNodeByID('Constant', -200, 0);
+      testController.moveNodeByID('Label', 230, 0);
+      testController.connectNodesByID(
+        'Constant',
+        'WidgetDropdown',
+        'Out',
+        'Select multiple',
+      );
+      testController.connectNodesByID('WidgetDropdown', 'Label', 'Out');
+      testController.executeNodeByID('Constant');
+    });
+    cy.wait(100);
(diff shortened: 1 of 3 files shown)


src/TestController.ts
```diff
@@ -39,14 +37,20 @@ export default class TestController {
...
+    node2Socket: string,
+  ) {
     const n1 = this.getNodeByID(node1ID);
     const n2 = this.getNodeByID(node2ID);
     const originSocket = n1.getOutputSocketByName(node1Socket);
-    PPGraph.currentGraph.connect(
-      originSocket,
-      n2.getSocketForNewConnection(originSocket),
-    );
+    const targetSocket =
```
magnificus, line 49: this is fine but could be made more elegant avoid undefined if we instead make another function that takes 4 inputs, then the function that takes 3 input creates the last argument via n2.getSocketForNewConnection(originSocket), and calls the function with 4 arguments

cypress/e2e/pnp/testWidgetNodes.cy.ts
```diff
@@ -0,0 +1,68 @@
+import { doWithTestController } from './helpers';
+
+describe('testWidgetNodes', () => {
```
magnificus, line 3: very nice with test

magnificus (approved):

src/TestController.ts
fakob: Sure I can do that, but if I understood correctly, would end up with lot's of duplicated code:
``` javascript
connectNodesByID(node1ID, node2ID, node1Socket, node2Socket) {
    const n1 = this.getNodeByID(node1ID);
    const n2 = this.getNodeByID(node2ID);
    const originSocket = n1.getOutputSocketByName(node1Socket);
    const targetSocket = n2.getInputSocketByName(node2Socket);

    PPGraph.currentGraph.connect(originSocket, targetSocket);
}

connectNodesByIDWithAutoTargetSocket(node1ID, node2ID, node1Socket) {
    const n1 = this.getNodeByID(node1ID);
    const n2 = this.getNodeByID(node2ID);
    const originSocket = n1.getOutputSocketByName(node1Socket);
    const targetSocket = n2.getSocketForNewConnection(originSocket);

    this.connectNodesByID(node1ID, node2ID, node1Socket, targetSocket);
}
```
@TobiasElinder better or did you mean something else?

magnificus: Ah, yes almost, just the difference is that the one with 3 arguments would call the function with 4 arguments, then you shouldnt have any copy pasted code, so

connectNodesByIDWithAutoTargetSocket
calls
connectNodesByID

by creating the last argument

fakob: That is what I do above :-)
The problem is that targetSocket for the one with 3 needs to be constructed.

magnificus: Hmm lets talk about this IRL today

magnificus: aha I understand what you are saying now, right yes fair enough lets do it as you did first.

PR #365, opened 2024-01-26
Added severity level to fix warning status not showing up

fakob:


src/classes/ErrorClass.ts
```diff
@@ -62,13 +78,17 @@ export abstract class PNPWarning extends PNPStatus {
 }
 
 export class PNPSuccess extends PNPStatus {
+  constructor(message?: string) {
+    super(message);
+    Object.setPrototypeOf(this, PNPSuccess.prototype);
+  }
+
   public isError(): boolean {
```
magnificus, line 86: I think this is a good change, but now we should remove this method, no longer relevant

magnificus (approved):
Good change, but lets remove old isError() method now that we have this new system

src/classes/ErrorClass.ts
fakob: Ah yes you are right. Will remove

PR #368, opened 2024-01-28
split db graph and graph meta into separate tables for performance + healthy refactoring, LOSES ALL OLD LOCAL PLAYGROUNDS!!!!!

magnificus:


src/PPStorage.tsx
```diff
@@ -177,57 +177,28 @@ export default class PPStorage {
...
   };
 
-  async downloadGraph(graphId = undefined) {
-    this.db
-      .transaction('rw', this.db.graphs, this.db.settings, async () => {
-        let serializedGraph;
-        let graphName;
-
-        const loadedGraphId = PPGraph.currentGraph.id;
+  async downloadGraph(graphId: string) {
```
fakob, line 180: With this change using the context menu download button or the create gist button will
* download/share the last *saved* version
* it will not include anything you might have changed, but not saved yet
* it will not allow you to download something you have not saved (including the remote graphs). you would always have to save them first

This is something I would not mind that much, if we had auto save. Or if it would let you know before downloading that you have unsaved changes. This does not mitigate the ability to save remote graphs which I guess we can solve differently.

I am not a big fan of this change.

fakob (approved):
Thanks for the db refactor. Regarding the download function change, I think we need a follow up or so as I don't find the behaviour ideal.

src/PPStorage.tsx
magnificus: OK will change it

magnificus: was not intending to change behaviour

PR #370, opened 2024-01-30
Fix issue when clicking on dynamic sockets

fakob:


src/classes/GraphClass.ts
```diff
@@ -47,6 +51,7 @@ export default class PPGraph {
   _showExecutionVisualisation: boolean;
   socketToInspect: null | PPSocket;
   selectedSourceSocket: null | PPSocket;
+  clickPoint: null | PIXI.Point;
```
magnificus, line 54: I think you can avoid saving this down as you already have the selectedSourceSocket, you can get the location from that, and we could say that we need to be X distance away from the socket, not from the click itself, I have the feeling it also might make it feel more natural with a distance from the socket rather than absolute distance from the origin click. Thoughts?

magnificus (approved):
Approve but added comment with suggestion

src/classes/GraphClass.ts
fakob: * the selectedSourceSocket refers to, as far as I can recall, the other end of the link. so we would need to save the additional selectedSourceTarget
* This is actually would I did first. Store the selectedSourceTarget to do the thing you talked about, use the actual socket as a click. However, this did not turn out to be great because
  * If you clicked on the edge of the socket, you would immediately reach the threshold and remove the link
  * If you clicked on the socket label, the link would also be removed 
The point of this is, did you intend to disconnect the socket and for this the indicator should be > how much did you move the mouse and not how far away of the center of the socket did you actually click.
Right now I find it to feel quite natural. I suggest to try this out and discuss again if you don't like it.

magnificus: OK, lets do it the way you propose

magnificus (approved):

PR #373, opened 2024-02-08
Refactor dialogs

fakob:
* The graph name is now selected when opening the edit dialog to rename it
* Moved the edit and delete dialog out of App.tsx
* Fixed an issue with the graph name not being update when renaming
* Added dialog tests for edit, delete and a one for share. No tests yet
  * for logging into github and sharing a gist
  * for download a file

Screenshot 2024-02-08 at 18 29 44


magnificus (approved):
looks like a good change

PR #374, opened 2024-02-10
Update enter behaviour for textfield

fakob:
This update only targets the string type
* Clicking Enter will
  * not create a line break anymore
  * trigger executeOptimizedChain
* The value is still updated on change, but the trigger allows a node which is not updated on change, to be triggered via enter. We can remove this, but it made sense to me, at least in Øysteins use case.
* For line break you can do shift+enter

### DIFF
--- a/src/widgets.tsx
+++ b/src/widgets.tsx
@@ -352,7 +352,6 @@ export const TextWidget: React.FunctionComponent<StringTypeProps> = (props) => {
   const handleKeyDown = (event) => {
     if (event.key === 'Enter' && !event.shiftKey) {
       event.preventDefault();
-      props.property.getNode().executeOptimizedChain();
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/widgets.tsx
```diff
@@ -343,6 +343,19 @@ export const TextWidget: React.FunctionComponent<StringTypeProps> = (props) => {
...
+  const handleChange = (event) => {
+    const value = event.target.value;
+    potentiallyUpdateSocketData(props.property, value);
+    setLoadedData(value);
+  };
+
+  const handleKeyDown = (event) => {
+    if (event.key === 'Enter' && !event.shiftKey) {
+      event.preventDefault();
+      props.property.getNode().executeOptimizedChain();
```
magnificus, line 355: I would make this only execute if node has update on change enabled

magnificus (approved):

src/widgets.tsx
fakob: Sounds good. I have removed the execute here completely, as we already have an onChange which triggers that there.

PR #375, opened 2024-02-13
Fix issue where remove link would remove all links

fakob:
* Fixed the issue where removing a link from a socket which has multiple links, would remove all of them
* Updated one of the tests to cover this case
* Renamed some variables


magnificus (approved):
Looks simpler (?) nice

PR #379, opened 2024-02-17
New welcome graph

fakob:
* Created new simple welcome graph
* Renamed all graphs to better work with the sentence - Plug and Playground lets you ...
* Added String node, which allows you to use all string functions - similar to math node
* Added Pulse node, which sends out a pulse - used to trigger an action on load
* Fixed an issue where the dataType configuration (e.g. min max values, etc.) were not loaded in configure
  * Added an exception for the dynamicEnumType, as it carries a function which can not be loaded
  * Added test


src/nodes/string.ts
```diff
@@ -0,0 +1,120 @@
...
+const parameterName2 = 'Parameter2';
+const outputName = 'Output';
+
+export class StringFunction extends PPNode {
+  constructor(name: string, customArgs: CustomArgs) {
+    super(name, {
+      ...customArgs,
+    });
+
+    this.onExecute = async function (input) {
```
magnificus, line 22: could this instead be defined as a normal function and not bound in constructor? Also could it be made to not set outputdata directly but instead return the output object in the execute onfunction?

src/nodes/math.ts
```diff
@@ -28,48 +18,39 @@ const singleNumberName = 'Number';
...
-        this.setOutputData('Output', Math[mathOption]());
-      } else if (staticMethodsWith2Parameters.includes(mathOption)) {
-        // check for staticMethodsWith2Parameters
-        this.setOutputData(
-          'Output',
-          Math[mathOption](input['Input'], input['Input2']),
-        );
-      } else {
-        this.setOutputData('Output', Math[mathOption](input['Input']));
+      const mathOption = input[inputOptionName];
```
magnificus, line 33: same as commented on stringtype, can we avoid binding this in constructor and just define overloaded function, and then also avoid setOutputData directly if possible

src/nodes/string.ts
```diff
@@ -0,0 +1,120 @@
...
+  getColor(): TRgba {
+    return TRgba.fromString(NODE_TYPE_COLOR.TRANSFORM);
+  }
+
+  protected getDefaultIO(): Socket[] {
+    const onOptionChange = (value) => {
+      this.setNodeName('String.' + value);
+    };
+    const str = Object.getOwnPropertyNames(String.prototype);
+    const strOptions = str
```
magnificus, line 73: if this could be extracted into its own function with a descriptive name it could increase readability.

src/nodes/datatypes/abstractType.tsx
```diff
@@ -114,6 +114,11 @@ export class AbstractType {
     return true;
   }
 
+  // load the saved dataType on configure
+  configureOnLoad(): boolean {
```
magnificus, line 118: what is this doing? :)

magnificus (changes requested):
Looks promising and good new graph but left a few comments

src/nodes/datatypes/abstractType.tsx
fakob: When we load a saved graph, we had for now, always loaded the default type ignoring any saved parameters.
`"dataType": "{\"class\":\"NumberType\",\"type\":{\"round\":true,\"minValue\":1,\"maxValue\":100,\"stepSize\":0.01}}",`
For example here in the NumberType. If you had changed and saved min and max values, they were ignored. I have fixed this now so we are not ignoring this any longer.
However, loading saved values does not make sense/work for the dynamicEnumType. I have therefore added this option so you can disable loading saved values on configure.
```
            if (dataType.configureOnLoad()) {
              matchingSocket.dataType = deSerializeType(item.dataType);
            }
```

src/nodes/string.ts
fakob: absolutely. thanks for pointing this out

src/nodes/datatypes/abstractType.tsx
magnificus: hmm OK

magnificus (approved):

PR #382, opened 2024-02-22
Fix draw memory leak

fakob:


magnificus (approved):
I had no idea they were not getting garbage collected

PR #383, opened 2024-02-25
Add html element output socket

fakob:
* Added an html element output socket, which outputs a reference to the html element which is displayed
  * Updated the graphs which had hardcoded this
* Added 2 more simple HTML nodes which just output a div/canvas
* Added 2 more examples using the observable plot library and the three.js library

### DIFF
--- a/src/nodes/draw/html.tsx
+++ b/src/nodes/draw/html.tsx
@@ -21,7 +21,7 @@ import HybridNode2 from '../../classes/HybridNode2';
 const inputSocketNameHeader = 'Header';
 const inputSocketNameHtml = 'Html';
 const reloadSocketName = 'Reload';
-const mainHtmlElementName = 'Main Html element';
+const mainHtmlElementName = 'HtmlElement';
 
 export class HtmlRenderer extends HybridNode2 {
   eventTarget: EventTarget;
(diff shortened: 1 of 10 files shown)


src/nodes/draw/html.tsx
```diff
@@ -179,13 +191,47 @@ interface MemoizedComponentProps {
...
     iframeRef,
     theme,
     headerData,
     htmlData,
     reload,
   }) {
+    const idWithDiv = `${id}-div`;
+
+    useEffect(() => {
+      if (iframeRef.current as any) {
```
magnificus, line 205: this whole segment looks quite finicky and side-effect-y, is it possible that we could make it simpler and more organized?

magnificus (approved):
Good with new graphs but the content of the function looks a bit messy IMO

src/nodes/draw/html.tsx
fakob: I had a simple and dirty setTimeout first, but wanted to do it properly :-)
The [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) was the thing I found which is what adds to the verbosity.
However, I now saw that the Frame npm package offers contentDidMount and contentDidUpdate callbacks which I switched it out with

magnificus: looks much cleaner now def

magnificus: looks cleaner definitely, is there a way we can maybe set this element somewhere else and retrieve it here? to avoid having to dig through all of this stuff with "childList" and outerElement

fakob: sorry, I saw your comment too late. Not sure I understand what you mean. Where should this be stored? in useState?

PR #385, opened 2024-02-28
Fix shift click to add to selection and updated test

fakob:
* Fix shift click to add to selection
* Updated test for it


magnificus (approved):
nice

PR #387, opened 2024-02-29
fix failing tests again

fakob:


fakob:
@TobiasElinder, I have used your tip of slowing down the windows machine with the power saving settings and did some trial and error. Now the tests should

magnificus (approved):
Thats good, but we still need the waits then? Pur async waiting thing doesnt work doe these?

fakob:
I don't think so. In this case for example, I am awaiting the async task to be finished, but the actual hybrid node rendering seems to take longer.
Screenshot 2024-02-29 at 22 56 26

PR #389, opened 2024-03-01
Add hash node

fakob:
* Added a node using the hash-wasm library
  * It supports a wide range of hash functions some of which need an object as input. As this needs some more knowledge, I have added a link to the documentation in the info.
  * for the HMAC function you can also define the hash method, and a key

https://github.com/fakob/plug-and-play/assets/4619772/83df2b98-fca3-49f1-920f-aba1031748ad

### DIFF
--- a/src/nodes/utility/utility.ts
+++ b/src/nodes/utility/utility.ts
@@ -341,11 +341,12 @@ export class LoadNPM extends CustomFunction {
   }
 }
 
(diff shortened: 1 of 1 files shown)


magnificus (approved):
nice, should add a method to get environmental variable back to PNP from companion, so that you can then bake it into a hash like this

fakob:
Great idea.

PR #391, opened 2024-03-06
real offset for draw combine array

magnificus:


src/nodes/draw/abstract.tsx
```diff
@@ -155,6 +155,18 @@ export abstract class DRAW_Base extends PPNode {
     offset: PIXI.Point,
   ): void;
 
+  private getContainer(
```
magnificus, line 158: this snuck in, not used yet, will be hopefully soon

magnificus: nvm i snuck the entire refactor in

magnificus:
This is a bit regression prone, I have tested it a bit and tests pass, but worth noting

src/nodes/draw/draw.tsx
```diff
@@ -462,37 +413,14 @@ export class DRAW_COMBINE_ARRAY extends DRAW_Interactive_Base {
...
-          executions,
-          x * spacingSize.width,
-          y * spacingSize.height,
-        );
-      }
-      shallowContainer.x = x * spacingSize.width;
-      shallowContainer.y = y * spacingSize.height;
-
-      /*
-      if (inputObject[objectsInteractive]) {
```
fakob, line 478: Do you think we can leave this in? It would be great to keep the interactivity here.

fakob (approved):

src/nodes/draw/draw.tsx
magnificus: we can keep it, but then we need to keep the container being created here, the hope was that spawning less container would make this less intensive (especially for the mac). I am fine with either.

fakob: Ah I see. I think the interactivity function is important, but there might be another way of achieving this. Instead of having an event listener on every subcontainer
* we add just one on the parent container and then use hitTest - https://pixijs.download/v6.1.0/docs/PIXI.InteractionManager.html#hitTest
* we would still make the sub objects = interactive, but then it would not need a separate container
Might be similar to what the pixi devtools do when you alt+click an element.

https://github.com/fakob/plug-and-play/assets/4619772/72f6fad3-e225-4a71-9b75-a553081051a2

What do you think?

magnificus: yes that looks like it could work, I will merge meanwhile as this does not change anything, but this sounds interesting

PR #397, opened 2024-03-16
Add margin and background to DRAW_Base

fakob:
* added a background and a margin to the DRAW_Base
  * the default is transparent and 0, but it allows to
  * easily add backgrounds and margins to any draw element like text, linegraph, combine, etc.
  * the margin is currently a single value for all sides, but I would like to support different margins in the future
* fixed an issue with anchor point. now everything rotates, scales and offsets correctly
* updated the reroute node so it can be moved a bit more easily
* updated the pivot example graph
* removed skew left over
* added `executions` undefined check due to this bug which I was not able to fix - https://trello.com/c/0Krxd8m0

https://github.com/fakob/plug-and-play/assets/4619772/60f9fa8b-6826-4f0d-b29f-bdf1643fd3a9

### DIFF
--- a/src/nodes/draw/abstract.tsx
+++ b/src/nodes/draw/abstract.tsx
@@ -11,17 +11,23 @@ import {
 } from '../../utils/constants';
 import { ArrayType } from '../datatypes/arrayType';
 import { BooleanType } from '../datatypes/booleanType';
+import { ColorType } from '../datatypes/colorType';
 import { DeferredPixiType } from '../datatypes/deferredPixiType';
 import { EnumType } from '../datatypes/enumType';
 import { NumberType } from '../datatypes/numberType';
 import { TwoDVectorType } from '../datatypes/twoDVectorType';
 import { TRgba } from '../../utils/interfaces';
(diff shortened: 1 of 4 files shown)


src/nodes/draw/abstract.tsx
```diff
@@ -301,23 +316,49 @@ export abstract class DRAW_Base extends PPNode {
...
+      pivotPoint.x * width - margin.left + myContainerBounds.x,
+      pivotPoint.y * height - margin.top + myContainerBounds.y,
     );
+
+    const bgColor = parseValueAndAttachWarnings(
+      this,
+      new ColorType(),
+      inputObject[bgColorName],
+    );
+    if (inputObject[marginSocketName] !== 0 || bgColor.alpha() !== 0) {
```
magnificus, line 350: maybe can check for > 0 instead of !== 0 
also can save to a boolean variable with a descriptive name to make it more readable (maybe hasValidBackgroundAndMargin)

magnificus (approved):

PR #398, opened 2024-03-18
disabled destroying of texture and basetexture to avoid image crash

magnificus:
dont know if this will bring back the leak we had....


fakob (approved):
I am pretty sure it will crash my machine again. I hope that pixi 8 will solve this. If not, maybe there is something else we can do about this issue.

PR #399, opened 2024-03-18
Improve combine node 2

fakob:
This is a temporary PR, as I am still looking for the issue with the last item missing the override. Maybe you can have a look through if you see something obviously wrong?


src/nodes/draw/draw.tsx
```diff
@@ -476,15 +610,11 @@ export class DRAW_COMBINE_ARRAY extends DRAW_Interactive_Base {
...
         );
       }
       if (inputObject[objectsInteractive]) {
-        addShallowContainerEventListeners(
-          shallowContainer,
-          this,
-          i,
-          executions,
-        );
+        addShallowContainerEventListeners(shallowContainer, this, i);
```
magnificus, line 617: missing executions here, intended???

src/nodes/draw/draw.tsx
```diff
@@ -344,14 +398,93 @@ export class DRAW_Combine extends DRAW_Base {
       positionKeys.reverse();
     }
 
+    const margin = {
```
magnificus, line 401: fairly certain something with this logic here is quacky, it looks very complex, im sure we can simplify it, this will likely also fix the bug

src/nodes/draw/abstract.tsx
```diff
@@ -352,12 +414,29 @@ export abstract class DRAW_Base extends PPNode {
...
       const background = new PIXI.Graphics();
       background.beginFill(bgColor.hex(), bgColor.alpha(true));
-      background.drawRect(
-        -margin.left + myContainerBounds.x,
-        -margin.top + myContainerBounds.y,
-        myContainerBounds.width + margin.left + margin.right,
-        myContainerBounds.height + margin.top + margin.bottom,
-      );
+
+      let bgX = -margin.left + myContainerBounds.x;
```
magnificus, line 418: this does not look good I think

src/nodes/draw/draw.tsx
fakob: executions is not used in the eventlistener, so I thought that I can remove it completely

fakob:
I have tried to keep the needed complexity completely within the combine node. Catering for this many options was challenging, but I think it is worth it.
* Updated the combine node and added the ability to
  * layout/position the inputs
    * manually (manual) - old behaviour
    * automatically separating them with a gap (gap)
    * automatically spreading them out between a fixed distance (spread)
  * set the direction of the layout (vertical or horizontal)
  * set the container size
    * based on the content size (hug)
    * based on the parent size (fill)
    * based on a fixed size (fixed)
  * align the inputs inside the container left/center/right, top/center/bottom
* Cleaned up abstract class
  * removed pivot (default pivot is now centered)
  * moved margin and background to combine node

Screenshot 2024-03-29 at 16 32 00

src/nodes/draw/abstract.tsx
```diff
@@ -275,13 +266,27 @@ export abstract class DRAW_Base extends PPNode {
     leading: false,
   });
 
-  private handleDrawing(drawingFunction: any): void {
+  private handleDrawing(drawingFunction: any, inputObject): void {
+    let passedInOverrideSettings;
+    // example for passing in overrides
```
magnificus, line 271: can remove

src/nodes/draw/abstract.tsx
```diff
@@ -275,13 +266,27 @@ export abstract class DRAW_Base extends PPNode {
     leading: false,
   });
 
-  private handleDrawing(drawingFunction: any): void {
+  private handleDrawing(drawingFunction: any, inputObject): void {
```
magnificus, line 269: inputobject not used? if so lets remove

src/nodes/draw/draw.tsx
```diff
@@ -344,16 +450,261 @@ export class DRAW_Combine extends DRAW_Base {
...
+
+    switch (widthBehaviour) {
+      case widthHeightFillOptions[0].text: // hug
+        predefinedBackgroundWidth = isVertical
+          ? groupWidthHeight.singleMaxWidth
+          : groupWidthHeight.netWidth +
+            spacingValue * (graphicsKeys.length - 1);
+        break;
+      case widthHeightFillOptions[1].text: // fill
+        if (parentBgWidth) {
```
magnificus, line 518: avoid nullcheck here, we should be able to say with confidence that it exists

src/nodes/draw/draw.tsx
magnificus: OK

src/nodes/draw/draw.tsx
```diff
@@ -344,16 +450,261 @@ export class DRAW_Combine extends DRAW_Base {
       positionKeys.reverse();
     }
 
-    graphicsKeys.forEach((graphicsKey, index) => {
-      const positionKey = positionKeys[index];
-      if (inputObject[positionKey]) {
-        const position = new PIXI.Point(
-          inputObject[positionKey].x,
-          inputObject[positionKey].y,
+    const bgColor = parseValueAndAttachWarnings(
```
magnificus, line 453: this function looks super verbose, can we simplify it? We should at least be able to split this function into several smaller ones.

magnificus (changes requested):
Looks promising and will maybe tie in well with the performance improvement (?) but left some comments

src/nodes/draw/draw.tsx
fakob: Well, the point is that `parentBgWidth` is only passed down/relevant when the parent is another combine node. However, the parent can be a multiplier node which does not pass this property down. In this case it just falls back to use the fixed with.

src/nodes/draw/abstract.tsx
fakob: will do

src/nodes/draw/abstract.tsx
fakob: will do

src/nodes/draw/draw.tsx
fakob: I had tried for some time to make it simpler, but due to its diverse functionality, I was not able to do that. I have now split certain parts into separate functions.

src/nodes/draw/draw.tsx
```diff
@@ -344,16 +572,203 @@ export class DRAW_Combine extends DRAW_Base {
...
+      } else {
+        // Calculate starting x based on horizontal alignment
+        switch (horizontalOption) {
+          case horizontalAlignmentOptions[0].text: // left
+            currentPositionX = padding.left;
+            break;
+          case horizontalAlignmentOptions[1].text: // center
+            currentPositionX =
+              padding.left +
+              (predefinedBackgroundWidth -
```
magnificus, line 689: same as comment above

src/nodes/draw/draw.tsx
```diff
@@ -344,16 +572,203 @@ export class DRAW_Combine extends DRAW_Base {
...
+          switch (inputObject[verticalAlignmentName]) {
+            case verticalAlignmentOptions[0].text: // top
+              currentPositionY = padding.top + bounds.height / 2;
+              break;
+            case verticalAlignmentOptions[1].text: // center
+              currentPositionY = padding.top + predefinedBackgroundHeight / 2;
+              break;
+            case verticalAlignmentOptions[2].text: // bottom
+              currentPositionY =
+                padding.top + predefinedBackgroundHeight - bounds.height / 2;
```
magnificus, line 738: one idea, is it possible to turn bounds or predefinedBackgroundHeight into something like a "heightToUse" varialble? then the difference could happen up there and dont need to diverge paths here

cypress/e2e/pnp/parallelizable/testCombineObjects.cy.ts
```diff
@@ -0,0 +1,464 @@
+import { doWithTestController, openNewGraph } from '../helpers';
```
magnificus, line 1: good with test

cypress/e2e/pnp/parallelizable/testCombineObjects.cy.ts
```diff
@@ -0,0 +1,464 @@
...
+    });
+    doWithTestController((testController) => {
+      testController.setNodeInputValue(
+        'DRAW_Combine-3',
+        'Direction',
+        'horizontal',
+      );
+      testController.executeNodeByID('DRAW_Combine-3');
+    });
+    cy.wait(100);
```
magnificus, line 187: are these waits needed? cant we use the await toast thing instead?

cypress/e2e/pnp/parallelizable/testCombineObjects.cy.ts
```diff
@@ -0,0 +1,464 @@
...
+        'fixed',
+      );
+      testController.setNodeInputValue(
+        'DRAW_Combine-3',
+        'Background height',
+        500,
+      );
+      testController.executeNodeByID('DRAW_Combine-3');
+    });
+    cy.wait(100);
```
magnificus, line 229: same as above and below etc

src/TestController.ts
```diff
@@ -50,6 +51,20 @@ export default class TestController {
...
+    const node = this.getNodeByID(id);
+    return findChildByName(node, name);
+  }
+
+  getCoordinatesOfChildren(container: PIXI.Container): number[][] {
+    const coordinates = container.children.map((child): number[] => [
+      child.x,
+      child.y,
+    ]);
+    console.log(coordinates);
```
magnificus, line 64: remove log

src/components/ContextMenus.tsx
```diff
@@ -610,17 +614,28 @@ export const NodeContextMenu = (props) => {
...
-            }}
-          >
-            <ListItemIcon>
-              <AddIcon fontSize="small" />
-            </ListItemIcon>
-            <ListItemText>Extract to Macro</ListItemText>
-          </MenuItem>
+        {areDrawNodesSelected && (
+          <>
+            <MenuItem
```
magnificus, line 619: do you want the extract macro thing to be related to this draw node specific thing? Possible typo

src/nodes/draw/draw.tsx
```diff
@@ -319,20 +319,252 @@ export class DRAW_Text extends DRAW_Base {
...
+  const computedBoundsObj = {
+    singleMaxWidth: 0,
+    singleMaxHeight: 0,
+    netWidth: 0,
+    netHeight: 0,
+  };
+  return graphicsKeys.reduce((acc, graphicsKey, index) => {
+    const drawingFunction: DeferredPixiTypeInterface = inputObject[graphicsKey];
+    const newHash = uuid();
+    drawingFunction.hash = () => newHash;
```
magnificus, line 373: drawingFunction.hash = uuid

src/nodes/draw/draw.tsx
```diff
@@ -319,20 +319,252 @@ export class DRAW_Text extends DRAW_Base {
...
+
+    const bounds = DRAW_Get_Bounds.getDrawingBounds(
```
magnificus, line 375: wondering, instead of setting the drawingfunction hash, you can just send in this UUID here into the getbrounds "callchain", a little cleaner.

src/nodes/draw/draw.tsx
```diff
@@ -319,20 +319,252 @@ export class DRAW_Text extends DRAW_Base {
   }
 }
 
+const bgColorName = 'Background color';
```
magnificus, line 322: Because of how large and complex this is, can we move into a separate file?

src/nodes/draw/draw.tsx
```diff
@@ -356,30 +589,220 @@ export class DRAW_Combine extends DRAW_Base {
...
+        } else {
+          currentPositionX += bounds.width + spacingValue;
+        }
+      });
+    }
+
+    const hasMarginOrBackground =
+      inputObject[paddingSocketName] > 0 || bgColor.alpha() > 0;
+
+    if (hasMarginOrBackground) {
```
magnificus, line 796: I will approve PR, but because of how complex this is I likely wont be able to bugfix it if it would break

src/pixi/utils-pixi.tsx
```diff
@@ -106,10 +106,17 @@ export const getNodesUpperLeftCorner = (
...
 
-export const getNodesBounds = (nodes: PPNode[]): PIXI.Rectangle => {
+export const getNodesBounds = (
+  nodes: PPNode[],
+  nodeBackgroundOnly = false,
+): PIXI.Rectangle => {
   let bounds = new PIXI.Rectangle();
   nodes.forEach((node: PIXI.DisplayObject, index: number) => {
-    const tempRect = node.getLocalBounds();
+    console.log(node.children[0].children[0]);
```
magnificus, line 115: I would not use this function this way it looks quite messy, if you want to do this I recommend just changing it in utlils where it is called instead of making this function less predictable.

cypress/e2e/pnp/parallelizable/testCombineObjects.cy.ts
fakob: I believe that I had issues when I ran `doWithTestController` back to back, but I can check again.

src/components/ContextMenus.tsx
fakob: Good catch! That is a mistake.

src/nodes/draw/draw.tsx
fakob: Ah yes.

src/nodes/draw/draw.tsx
fakob: The getDrawingBounds function uses the hash for lookup and not the callchain. I am not fully sure how your logic works, but I believe it did not update when I passed in the UUID as callchain.

src/nodes/draw/draw.tsx
fakob: Good idea. Will do.

src/pixi/utils-pixi.tsx
fakob: Ok, can do.

src/nodes/draw/draw.tsx
fakob: Hmm, actually that ended up in an error inside DRAW_Get_Bounds.getDrawingBounds
> TypeError: Cannot read properties of undefined (reading '×')
As I only need to set a unique value once. I kept it.

cypress/e2e/pnp/parallelizable/testCombineObjects.cy.ts
fakob: So it seems that I do need a short wait after every`executeNodeByID` call.

PR #400, opened 2024-03-20
removed custom selection logic for macro, it wasnt working well

magnificus:


fakob (approved):
I am not super familiar how the macro selection worked. Should we then also remove the corresponding functions on the NodeClass or do you want to build on them again later?

magnificus:
removing them is a good idea probably yes, just wasnt sure if they would be useful at some point

PR #401, opened 2024-03-31
Draw caching for performance

magnificus:
I need to do some more testing, this is very regression prone, but there is a significant (and should be very noticeable) performance improvement when using any form of composite drawing


fakob (approved):
Good idea to cache the containers. Do I see this correctly, that we are adding another container layer, or is this a substitute for one in the abstract class?

PR #402, opened 2024-04-07
small refactor and fix for twodvectortype not updating

magnificus:


src/nodes/datatypes/twoDVectorType.tsx
```diff
@@ -9,6 +9,11 @@ export interface TwoDVectorTypeProps extends DataTypeProps {
   dataType: TwoDVectorType;
 }
 
+export interface TwoDVectorTypeInterface {
+  x: number,
+  y: number
+}
+
```
fakob, line 16: Maybe consolidate with TTwoDVector type in interfaces.tsx as they seem to be the same

src/widgets.tsx
```diff
@@ -825,25 +826,49 @@ export const ColorWidget: React.FunctionComponent<ColorTypeProps> = (props) => {
...
+}
+
 export const TwoDNumberWidget: React.FunctionComponent<DataTypeProps> = (
-  props,
+  props: TwoDVectorTypeProps,
 ) => {
-  const propertyX = props.property.data?.x || 0;
-  const propertyY = props.property.data?.y || 0;
+  let incoming = { x: props.property.data.x, y: props.property.data.y };
+  const [current, setCurrent] = useState(incoming);
```
fakob, line 857: If we are removing this fallback to 0 if there is no x or y data, maybe we should add a custom parser? Otherwise we might get an error when changing data type from a one dimensional type? Or am I misunderstanding this?

fakob (approved):

src/nodes/datatypes/twoDVectorType.tsx
magnificus: OK will do

src/widgets.tsx
fakob: Makes sense to do this outside. Thanks!

fakob: Makes sense to do this outside. Thanks!

magnificus: I found that it was already doing this in the parse method

PR #405, opened 2024-04-19
Add dropdown alternative and removed made all colors use TRgba

fakob:
* Added an option in the enumType to display the options as toggle buttons. This has the advantage of showing all options right away and only needing one click to change instead of 2 with a dropdown
  * Updated some of the node sockets where the options were not too many. For example Draw Shape, Http, Combine Object
  * However the component does support horizontal scrolling if there are too many options
* Removed all other references to the `Color` package. All colors are now using the TRgba class

Screenshot 2024-04-19 at 18 47 36


magnificus (approved):
very nice

PR #406, opened 2024-04-19
Add pan when mouse dragging close to edge

fakob:
* activated one of the pixi-viewport plugins I had forgotten about which pans the viewport
  * ~~when moving the mouse close to the edge~~
  * when dragging nodes or connections. It can be a bit choppy

https://github.com/fakob/plug-and-play/assets/4619772/7bb4a15a-575b-418f-a260-655379b537ca


fakob:
Actually, let me pull that back for a bit. I noticed that this feature should only be active when the mouse is dragging a node, a selection or a connection and not also when the mouse is just close to the border.

fakob:
The pan behaviour is now only active when clicking/dragging
* a Node
* a selection
* a canvas

magnificus (approved):

